PR #6, opened 2021-05-01
Fix package.json scripts

chapel:
No need to specify the bins directly in the script configs can just list them directly. This ensures that both they are portable to all systems (windows/mac/linux) but also more flexible. More info here: https://docs.npmjs.com/cli/v7/using-npm/scripts#path


HPaulson (approved):

lazynina (approved):
Thanks for the contribution!

PR #23, opened 2021-05-04
Ln/link diamond notification to post

lazynina:


src/app/right-bar-creators/right-bar-creators.component.html
```diff
@@ -3,7 +3,7 @@
   <search-bar></search-bar>
 
   <div class="pt-10px"></div>
-
+a
```
maebeam, line 6: rm

src/app/notifications-page/notifications-list/notifications-list.component.ts
```diff
@@ -219,12 +219,27 @@ export class NotificationsListComponent implements OnInit {
       const isDiamond = cctMeta.DiamondLevel > 0;
       result.icon = isDiamond ? "icon-diamond fc-blue" : "fas fa-paper-plane fc-blue";
       const diamondText = isDiamond
```
maebeam, line 221: Can we make what one giant if else block for `isDiamond { result.action = '' } else { result.action = '' }` now that there appears to be no overlap between these two?

maebeam (changes requested):

src/app/notifications-page/notifications-list/notifications-list.component.ts
```diff
@@ -216,15 +216,29 @@ export class NotificationsListComponent implements OnInit {
...
-      )} ${cctMeta.CreatorUsername} coins`;
+      if (cctMeta.DiamondLevel) {
+        result.icon = "icon-diamond fc-blue";
+        let postText = "";
+        if (cctMeta.PostHashHex) {
+          const post = this.postMap[cctMeta.PostHashHex];
+          if (!post) {
+            return null;
+          }
+          const truncatedPost = _.truncate(_.escape(`${post.Body} ${post.ImageURLs?.[0] || ""}`));
```
maebeam, line 227: can we move this to a method to avoid code duplication?

looks great otherwise!

PR #34, opened 2021-05-06
fix: embed using the no cookie version of youtube iframes and vimeo

chafreaky:
This is to avoid embedded youtube videos from dropping cookies on users when viewing them. See here: https://support.google.com/youtube/answer/171780?hl=en-GB#zippy=%2Cturn-on-privacy-enhanced-mode

For vimeo, passing along a `?dnt=1` flag prevents dropping of cookies when creating iframe


chafreaky:
If the implementation of this is okay with the core devs I can also add dnt flags for Vimeo (I don't think tiktok has one?)

chafreaky:
Found surfaces for improvement, moving back to draft

lazynina (changes requested):
This looks like a good change, but we need to update the `isYoutubeFromURL` function to recognize `youtube-nocookie.com` as a valid hostname for youtube and update the `youtubeParser` function to recognize it as a valid youtube URL in the event a user places a url of the no-cookie format into the embed input element.

chafreaky:
Sure, I can make those changes. 

I left it as is because most users don't know of the existence of the youtube-nocookie.com domain. It actually doesn't work unless used in an iframe. 

I'll update the PR, thanks @lazynina!

lazynina:
> Sure, I can make those changes.
> 
> I left it as is because most users don't know of the existence of the youtube-nocookie.com domain. It actually doesn't work unless used in an iframe.
> 
> I'll update the PR, thanks @lazynina!

fine to leave it out from regex since it's only used for iframes

lazynina:
> I thought about this some more and I don't think it's a good idea to change those two functions @lazynina.
> 
> The goal of this PR is to still accept a vanilla Youtube URL as you initially intended but use the `nocookie` version when building the iframe. The user has no knowledge of this happening.
> If we allowed the `nocookie` version of the URL to be an input then you'd have to accept completely different URL paths (the nocookie version is _only_ for embedding, it's not user accessible).
> 
> On top of this, with the current implementation, adding a flag to the `constructYoutubeEmbedURL` function would allow people to choose whether or not they want to be tracked and therefore build the iframe URL differently while still accepting the same input URL. Something other node frontends may want to add.
> 
> Thoughts?

You need to at least update `isYoutubeFromURL` as youtube embed do not appear in posts with your current implementation.  We check that  video URLs match on of the whitelisted URL formats in `getEmbedVideoURL` and if it does not match, we return a falsy value. Alternatively, you could create a new function `isYoutubeNoCookieFromURL` to match this specific case and add an if statement in `getEmbedVideoURL` to handle it. 

Adding a flag to the `constructEmbedURL` makes sense.

lazynina (changes requested):
see comments above for requested changes.

chafreaky:
Yep - will adjust, thank you @lazynina!

PR #35, opened 2021-05-06
fix unread box when >100 unread messages

tijno:
I noticed this when re-syncing my node.

formatting-issue

This PR removes the width and adds some padding to fix this.

fix-100+

Also still ok with smaller unread messages.

20210506-ypo9yNCH

Small change but hopefully useful.

### DIFF
--- a/src/app/bottom-bar-mobile/bottom-bar-mobile.component.html
+++ b/src/app/bottom-bar-mobile/bottom-bar-mobile.component.html
@@ -34,7 +34,7 @@
           *ngIf="globalVars.messageNotificationCount > 0"
           class="bottom-bar__message-notifications-bubble roboto-mono"
         >
-        {{ globalVars.messageNotificationCount > 999 ? "loads" : globalVars.messageNotificationCount }}
+        {{ globalVars.messageNotificationCount > 999 ? "999+" : globalVars.messageNotificationCount }}
         </div>
       </i>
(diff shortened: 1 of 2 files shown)


chafreaky (commented):
LGTM! Doesn't handle large numbers but I doubt we'd want to cater for inboxes with 1m+ messages.

One thing though, there is an adjustment to be made on the unread messages pill for the mobile view. It's missing a right alignment. As of now it goes off view if a large number is written. Not sure if you wanted to adjust that as well.

tijno:
Good point ... but that would require me to fake the unreads as I cleared them all. Let me see what I can do.

tijno:
@chafreaky seems ok on mine - could you show maybe or suggest change ?

20210506-45koBEPP
20210506-puFCmA1X

chafreaky:
It's not a big concern it's just my OCD trying to get things right, but here is an example below (Using IPhone X viewport on Chrome):

Screenshot 2021-05-06 at 15 06 14

tijno:
maybe this instead?

20210506-x8EZ0Oy9

chafreaky:
Hahahaha, love that!

maebeam:
Sorry to nit but I don't like "loads." Can we do 999+ or something more user friendly?

tijno:
@maebeam thats done - changed to 999+ for left and bottom bars

PR #36, opened 2021-05-06
add button to mark all messages as read

tijno:
Another issue I ran into using my own node - messages show up as unread and there was no easy way to mark them all as read.

fix-100+

So added a button. Im not familiar with ts/angular so im sure several mistakes made and improvements possible. 

Would be good if someone with better knowledge & experience could chip in with improvements?

When you have unread messages - the button is active like this:

unread-messages-button-active

When all are read, or button is pressed, it is inactive and gray.

all-read-inactive-button

I dont really like the icon but I could not figure out how to reliably add a suitable read/unread icon from fontello as adding to CSS did not seem enough.

### DIFF
--- a/src/app/messages-page/messages-inbox/messages-inbox.component.html
+++ b/src/app/messages-page/messages-inbox/messages-inbox.component.html
@@ -1,9 +1,10 @@
 <!-- Top Bar -->
-<div
-  *ngIf="!isMobile"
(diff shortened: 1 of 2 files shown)


bluepartyhat:
This design makes a lot of sense to me. I'm currently working on making messages more robust on the backend, but this will mesh well with that when combined.

tijno:
hopefully that will be out soon. Do you want me to resubmit the PR when you are done, or will you include a mark all as read function @bluepartyhat ?

PR #38, opened 2021-05-07
support for embedding videos and images in quoted reclouts

lazynina:


dgsus (changes requested):
the last "< / div >" on src/app/feed/feed-create-post/feed-create-post.component.html is breaking the code.

PR #42, opened 2021-05-07
Creator Coin Significant Figures Rationalization

LukeWalsh:
I use a slightly nonstandard approach with my creator coin in a 99.99 founder percentage. This surfaced an issue for visualizing holders of my coin which has an easy fix. I could also imagine the reverse case being important, my holders viewing their correct coin balance in the wallet.

This pull request shows example changes which would allow for the high creator coin percentage.

In making these changes I noticed that sometimes the creator coin significant figures are calculated through a global function, but, on the profile page this function was not used. This led me to think that the global balance formatting function should be used everywhere & the second numeric input should be removed.

In the meantime these changes could be simply merged as-is to support the rationalization of significant figures with a separate discussion about refactoring the approach to formatting.

### DIFF
--- a/src/app/creator-profile-page/creator-profile-hodlers/creator-profile-hodlers.component.html
+++ b/src/app/creator-profile-page/creator-profile-hodlers/creator-profile-hodlers.component.html
@@ -28,7 +28,7 @@
       </a>
     </div>
     <div class="col-3 d-flex align-items-center mb-0">
-      {{ (row.BalanceNanos / 1e9).toFixed(4) }}
+      {{ (row.BalanceNanos / 1e9).toFixed(9) }}
     </div>
     <div class="col-3 d-flex align-items-center mb-0">
       ≈ {{ globalVars.creatorCoinNanosToUSDNaive(row.BalanceNanos, profile.CoinPriceBitCloutNanos, true) }}
--- a/src/app/wallet/wallet.component.html
+++ b/src/app/wallet/wallet.component.html
@@ -168,7 +168,7 @@
                     }}
                   </div>
                   <div class="text-grey8A fs-12px text-right">
-                    {{ globalVars.nanosToBitClout(creator.BalanceNanos, 4) }}
+                    {{ globalVars.nanosToBitClout(creator.BalanceNanos, 9) }}
                   </div>
                 </div>
               </div>


maebeam:
9 sig figs is unnecessary. I'm open to considering a bump to 5 but even that seems excessive.

LukeWalsh:
Hi Mae, yes, I'm basically trying to get to 4-5 significant figures displayed in the case of my creator coin (check my profile for an example of how this plays out for high founder ownership %). The current implementation of f.nanosToBitClout actually doesn't treat the second input as significant figures today.

> For a decimal number that is less than 1, all figures are significant except the leading zeros after its decimal point.  For example 0.235kg is less than 1 and has 3 sig. figs.  This is equivalent to 235 grams.  Suppose instead we have 0.000235kg.  The three zeros before 235 but after the decimal point do not count as significant.  This is equivalent to 235milli-grams and has 3 sig. figs., again.

Overall I think that **f.nanosToBitClout should be modified to handle the formatting here & in any future display areas without taking a second numeric input** (and we could possibly remove fractions entirely such that we express the number of nano creator coins with metric prefixes instead). This would lead to two possible outcomes if we assumed holding 4 significant figures as the constant (but actually expressing 4 significant figures in the case of zeros after the decimal point unlike the current implementation):

### Option A: Express creator coins with nanos becoming 'a coin: c'
So, someone who owns ~1 WHOLE creator coin might see: '1.235Gc'
_--> The common case today is around here, between these two_
Someone who owns 1/1000th of '1.235G' would see: '1.235Mc'
Someone who owns 1/1000th of '1.235M' would see: '1.235kc'
Someone who owns ~1/10th of '1.235k' would  see: '123c' _(without +0.5 since they could not own fractional nanos & this is the base unit)_

### Option B: Keep the nanos idea as-is and use metric prefixes for the highly fractionalized case
So, someone who owns 1 WHOLE creator coin might see: '1.235c'
_--> The common case today is around here, between these two_
Someone who owns 1/1000th of '1.235c' would see: '1.235mc'
Someone who owns 1/1000th of '1.235mc' would see: '1.235μc'
Someone who owns 1/10th of '1.235mc' would see: '1.235nc'  _(without +0.5 since they could not own fractional nanos & this is the base unit)._

Does that make sense? I'm sure there are other Options for handling this as well, and decisions within A/B around when to use the SI prefix if there was an idea to do something slightly less scientifically standard, but, more immediately understandable as a transition from where we are today.

maebeam:
Thanks for covering all the options. For now we're going to keep all units in "bitclout" and not introduce the concept of a smaller denomination. I also don't see much value in rendering more than four digits in these places because most people own at least $1 of coins they care about.
 
The "sig fig" terminology is used incorrectly here. It's actually means "max number of decimals" and I'm open to renaming it.

LukeWalsh:
Got it, thanks for the feedback! I won't submit a pull request for the refactor in that case.

PR #45, opened 2021-05-07
use router for internal link if host matches

tijno:
This is a small enhanement to user experience when clicking on posts linking to node links, like bitclout posts. Right now these always open in a new tab, and require the "tap to load page" action. Instead they could use the router to speed up loading, and reduce system strain.

Here is what it looks like with the PR implemented (in this example on localhost of course due to development)

routed-link

Autolinker allows for custom function to run on matched link patterns. I used this to get links to open with router rather then in target=_blank.

Hopefully this is suitable.

Let me know if you want me to make any changes.


tijno:
One consideration is whether nodes should route even when url is bitclout.com - reverse of course is harder but could be possible for certain urls that look like bitclout urls

maebeam:
This feels a bit hacky. I'm also planning on ripping out these autolinker classes soon anyways.

tijno:
what would make it less hacky? just so i know for future prs

tijno:
As this has not been adopted ill remove it.

PR #46, opened 2021-05-07
Prevent screen overflow on search dropdown

chafreaky:
Prevents behavior seen below that makes the search dropdown overflow viewport:

Screenshot 2021-05-07 at 18 10 39


maebeam:
Can we make the max height like 600px or something longer?

Also, before/after screenshots are helpful for design PRs

chafreaky:
Changed! See screenshot below for after new style:

Screenshot 2021-05-08 at 13 03 21

maebeam:
Thank you!

PR #66, opened 2021-05-19
add follow btn back to follow feed

lazynina:


maebeam:
How come? There’s not enough space on the feed and unfollow should be a fairly rare action

PR #67, opened 2021-05-21
Ln/right sidebar enhancements

lazynina:


src/lib/services/pulse/pulse-service.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  getProfilesForPulseLeaderboard(
+    res: PulseLeaderboardResponse,
+    leaderboardType: PulseLeaderboardType
+  ): Observable<any> {
+    const results = res.results;
+    return forkJoin(
+      results.map((result) =>
+        this.backendApi.GetSingleProfile(this.globalVars.localNode, result.public_key, null).pipe(
```
maebeam, line 73: Please do one single `GetUsersStateless` call here

src/app/right-bar-creators/right-bar-creators.component.ts
```diff
@@ -1,11 +1,70 @@
...
+    [RightBarCreatorsComponent.COMMUNITY.name]: RightBarCreatorsComponent.COMMUNITY,
+  };
+
+  ngOnInit() {
+    const defaultTab = this.backendApi.GetStorage(RightBarCreatorsComponent.RightBarTabKey);
+    this.activeTab = defaultTab in this.chartMap ? defaultTab : this.selectRandomTab();
+    this.selectTab(true);
+  }
+
+  selectRandomTab(): string {
```
maebeam, line 57: Why? Can we just decide on a default?

src/app/right-bar-creators/right-bar-creators-leaderboard/right-bar-creators-leaderboard.component.html
```diff
@@ -1,26 +1,26 @@
 <a
-  *ngFor="let profileEntryResponse of globalVars.rightBarLeaderboard"
-  [routerLink]="['/' + globalVars.RouteNames.USER_PREFIX, profileEntryResponse.Username]"
+  *ngFor="let profileEntryResponse of getActiveTabLeaderboard()"
```
maebeam, line 2: this was stored in GlobalVars so that we don't refetch top creators whenever the user navigates to a new page. also, a variable is better than a method because whenever angular decides to re-render the page the result is cached

src/app/right-bar-creators/right-bar-creators-leaderboard/right-bar-creators-leaderboard.component.ts
```diff
@@ -1,57 +1,84 @@
...
-      .subscribe(
-        (response) => {
-          this.globalVars.rightBarLeaderboard = response.ProfilesFound.slice(
-            0,
-            RightBarCreatorsLeaderboardComponent.MAX_PROFILE_ENTRIES
-          );
+    const pulseService = new PulseService(this.httpClient, this.backendApi, this.globalVars);
+
+    pulseService.getBitCloutLockedLeaderboard().subscribe((res) => {
+      pulseService.getProfilesForPulseLeaderboard(res, PulseLeaderboardType.BitCloutLocked).subscribe({
```
maebeam, line 45: the response from `getBitCloutLockedLeaderboard` should include the profiles fetched from GetUsersStateless by chaining promises together. there are examples of this in backend service.

src/app/right-bar-creators/right-bar-creators.component.ts
lazynina: @diamondhands0 requested a randomization until user selects a tab.  I'd prefer to pick a default -- I think "Top Gainers" would be the logical default.

PR #69, opened 2021-05-22
cdd/Fix embed validation

connerdouglass:
In the `VideoUrlParserService` class, there were a few places where validation wasn't specific enough.

Domains such as "wwwzyoutube.com" and "nottiktok.com" would have previously passed validation. On clients with a strong CSP, these are blocked by the browser. However, clients not configured with a CSP would possibly be open to injection of malicious iframes.

This PR improves the regular expressions used for validation, and adds some test cases (closer to what an adversary or penetration tester would try to input).


connerdouglass:
@chafreaky and @lazynina, just a heads up that this change may conflict with PR #34, depending on if/when these get merged.

lazynina (approved):

PR #78, opened 2021-05-28
fade text if longer then 280

tijno:
The feed is starting to get really hard to read sometimes with people using Cloutfeed and other tools to post long long posts with way more then the ~~260~~ 280 Max characters on the main node.

So I made this small amendment that fades out and hides body overflow on a feedpost if it is longer then ~~260~~ 280 chars.

Made sure its not hidden when viewing the actual post page, instead of feed.

Could be enhanced by adding a "read more" button, but i left that out for now. Left todo for that in comment.

### DIFF
--- a/src/app/feed/feed-post/feed-post.component.html
+++ b/src/app/feed/feed-post/feed-post.component.html
@@ -214,9 +214,10 @@
         </div>
 
         <!-- Content -->
+        <!-- TODO: Add a read more button for posts that are too long? -->
         <div
           class="roboto-regular mt-1"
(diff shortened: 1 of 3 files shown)


iPaulPro:
Why 260 rather than the original 280?

tijno:
Need to double check I thought it was 260 @iPaulPro

tijno:
Ah warning shows at 260 - ill amend to 280

maebeam:
Thanks for the suggestion! For now, we don't mind users making longer form posts. We may revisit this in the future.

PR #79, opened 2021-05-29
add-amount-to-notification

tijno:
Lots of users mentioned it would be nice to see the amount in the basic transfer notification when someone sends you bitclout. Im finding this too. So I added this small change which shows the transferred bitclout in the notification.


tijno:
This does not handle UTXO transactions so cant get actual amount in all circumstances

PR #81, opened 2021-05-29
Rph/interaction-details

redpartyhat:


src/app/feed/feed-post/feed-post.component.html
```diff
@@ -335,4 +334,59 @@
...
+      class="mt-2px"
+      *ngIf="showIconRow"
+      [noMaxWidth]="true"
+      [post]="post"
+      [postContent]="postContent"
+      [parentPost]="parentPost"
+      [hideNumbers]="true"
+      [afterCommentCreatedCallback]="afterCommentCreatedCallback"
+      [afterRecloutCreatedCallback]="afterRecloutCreatedCallback"
+    ></feed-post-icon-row>
```
lazynina, line 390: The width on this component looks a bit strange IMO.  I would expect the icons in the "main" post to line up with the icons of any parents or children.

src/app/feed/feed-post/feed-post.component.ts
```diff
@@ -159,6 +171,46 @@ export class FeedPostComponent implements OnInit {
...
+
+  openQuoteRecloutsModal(event, imageURL) {
+    event.stopPropagation();
+    this.modalService.show(QuoteRecloutsModalComponent, {
+      class: "modal-dialog-centered",
+      initialState: {
+        postHashHex: this.post.PostHashHex,
+      },
+    });
+  }
```
lazynina, line 212: Doesn't look like we use `imageURL` in any of these.  Also seems like we can refactor since the only thing that changes is the component passed in. 

```
openInteractionModal(event, component): void {
    event.stopPropagation();
    this.modalService.show(component, {
      class: "modal-dialog-centered",
      initialState: {
        postHashHex: this.post.PostHashHex,
      },
    });
}
```

and then you could simplify each of these, for example:
```
openQuoteRecloutsModal(event): void {
    this.openInteractionModal(event, QuoteRecloutsModalComponent);
}
```

src/app/infinite-scroller.ts
```diff
@@ -0,0 +1,58 @@
...
+          pageResults = pageResults.reduce((acc, result) => [...acc, ...result], []);
+          const start = startIdx - startPage * this.pageSize;
+          const end = start + endIdx - startIdx + 1;
+          return pageResults.slice(start, end);
+        });
+      },
+      settings: {
+        startIndex: 0,
+        minIndex: 0,
+        bufferSize: 50,
```
lazynina, line 53: Can we make bufferSize an input to the constructor? I could see this value changing depending upon what type of content we're scrolling. Can set it to default to 50

src/app/likes-modal/likes-modal.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.loading = true;
+    return this.backendApi.
+      GetLikesForPost(
+        this.globalVars.localNode, 
+        this.postHashHex, 
+        this.pageOffset, 
+        this.pageSize, 
+        this.globalVars.loggedInUser.PublicKeyBase58Check)
+      .toPromise()
+      .then((res) => {
```
lazynina, line 51: Can we use a subscription instead of promises?

lazynina (changes requested):

lazynina (approved):
:+1:

PR #90, opened 2021-06-01
Add messages notifications for most recent tab.

bluepartyhat:


src/app/global-vars.service.ts
```diff
@@ -156,35 +157,87 @@ export class GlobalVarsService {
...
+      this.messagesRequestsHoldingsOnly = false;
+      this.messagesRequestsFollowersOnly = false;
+      this.messagesRequestsFollowedOnly = false;
+      this.messagesSortAlgorithm = "time";
+    } else if (tabName == "Custom") {
+      this.messagesRequestsHoldersOnly = this.backendApi.GetStorage("customMessagesRequestsHoldersOnly");
+      this.messagesRequestsHoldingsOnly = this.backendApi.GetStorage("customMessagesRequestsHoldingsOnly");
+      this.messagesRequestsFollowersOnly = this.backendApi.GetStorage("customMessagesRequestsFollowersOnly");
+      this.messagesRequestsFollowedOnly = this.backendApi.GetStorage("customMessagesRequestsFollowedOnly");
+      this.messagesSortAlgorithm = this.backendApi.GetStorage("customMessagesSortAlgorithm");
```
lazynina, line 203: this conditional logic could be simplified:
```
if (tabName !== "Custom) {
      this.messagesRequestsHoldingsOnly = tabName === "My Holders";
      this.messagesRequestsFollowersOnly = false;
      this.messagesRequestsFollowedOnly = false;
      this.messagesSortAlgorithm = "time";
} else {
... keep all the same logic for the custom case.
}
```

src/app/bottom-bar-mobile/bottom-bar-mobile.component.html
```diff
@@ -31,10 +31,10 @@
     >
       <i class="icon-messages fs-25px">
         <div
-          *ngIf="globalVars.messageNotificationCount > 0"
+          *ngIf="globalVars.messageResponse && globalVars.messageResponse.NumberOfUnreadThreads != 0"
```
lazynina, line 34: `*ngIf="globalVars.messageResponse?.NumberOfUnreadThreads"`

src/app/global-vars.service.ts
```diff
@@ -156,35 +157,87 @@ export class GlobalVarsService {
...
+      storedTab = "My Holders";
+      this.backendApi.SetStorage("mostRecentMessagesTab", storedTab);
+    }
+
+    // Set the filters most recently used and load the messages
+    this._setMessagesFilter(storedTab);
+    this._loadInitialMessages();
+  }
+
+  _setMessagesFilter(tabName: any) {
```
lazynina, line 183: by convention, function names that begin with an underscore are considered private.  Since this function is used outside of this component, let's remove the underscore.

src/app/global-vars.service.ts
```diff
@@ -156,35 +157,87 @@ export class GlobalVarsService {
...
+    } else if (tabName == "Custom") {
+      this.messagesRequestsHoldersOnly = this.backendApi.GetStorage("customMessagesRequestsHoldersOnly");
+      this.messagesRequestsHoldingsOnly = this.backendApi.GetStorage("customMessagesRequestsHoldingsOnly");
+      this.messagesRequestsFollowersOnly = this.backendApi.GetStorage("customMessagesRequestsFollowersOnly");
+      this.messagesRequestsFollowedOnly = this.backendApi.GetStorage("customMessagesRequestsFollowedOnly");
+      this.messagesSortAlgorithm = this.backendApi.GetStorage("customMessagesSortAlgorithm");
+    }
+  }
+
+  _loadInitialMessages() {
```
lazynina, line 207: see my comment above about js function naming conventions

src/app/global-vars.service.ts
lazynina: note: we haven't been great about this, but would like to try to be more consistent going forward.

lazynina (changes requested):
some stylistic changes and one edge case:
```
If I open the messages tab and the default thread I'm on is "unread", it remains unread after I navigate away from messages. Even if I respond to the message, it remains unread. Seems like we only mark things as read when we click on a thread from the thread view
```

lazynina (approved):

PR #95, opened 2021-06-06
Added Dynamic Profile Username Titles and Titles for all pages

carsenk:
This gives the feature of titles on pages for BitClout vs. the old every page being "Welcome to BitClout"

This also gives dynamic profile username pages so for example my https://bitclout.com/u/carsenk will now have the title of my creator coin/username

Example preview:
IqscY0GCo8

I will be adding a PR next for dynamic Twitter OG Metatags so that BitClout posts and pages show up on social media sites similar to how other modern sites or aka Twitter does.

💎🤲💎🤲💎🤲

### DIFF
--- a/src/app/admin/admin.component.html
+++ b/src/app/admin/admin.component.html
@@ -1,4 +1,5 @@
 <div *ngIf="globalVars.nodeInfo != null">
+  {{ setTitle('Admin - BitClout') }}
   <!-- Top Bar -->
   <div
     *ngIf="!isMobile"
--- a/src/app/creator-profile-page/creator-profile-details/creator-profile-details.component.ts
+++ b/src/app/creator-profile-page/creator-profile-details/creator-profile-details.component.ts
@@ -5,7 +5,7 @@ import { ActivatedRoute, Router } from "@angular/router";
 import { Location } from "@angular/common";
 import { SwalHelper } from "../../../lib/helpers/swal-helper";
 import { CreatorProfileTopCardComponent } from "../creator-profile-top-card/creator-profile-top-card.component";
-import { Title } from '@angular/platform-browser';
+import { Title, Meta } from '@angular/platform-browser';
 
 @Component({
   selector: "creator-profile-details",
@@ -44,6 +44,7 @@ export class CreatorProfileDetailsComponent {
     private router: Router,
     private location: Location,
     private titleService: Title,
+    private metaService: Meta,
   ) {
     this.route.params.subscribe((params) => {
       this.userName = params.username;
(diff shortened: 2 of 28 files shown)


carsenk:
I just added dynamic metatags as well to this PR :)

Dynamic Metatags on all Creator Coin pages

Dynamic Metatags and Titles on all Posts

image

image

Ideally end result once on the main domain will be things like this on Discord/Twitter/etc.

image

tijno:
Nice work on the titles - i often have multiple bitclout tabs open this will help a lot.

Though not sure it will help with link previews in discord / twitter cards.

Dont they just make single get request to the url, dont process the JS, and would therefore receive the non JS rendered html?

Here is example of the request discord makes. 

image

So for link previews to work the server would need to return the html with the metatags already included.

carsenk:
> Nice work on the titles - i often have multiple bitclout tabs open this will help a lot.
> 
> Though not sure it will help with link previews in discord / twitter cards.
> 
> Dont they just make single get request to the url, dont process the JS, and would therefore receive the non JS rendered html?
> 
> Here is example of the request discord makes.
> 
> image
> 
> So for link previews to work the server would need to return the html with the metatags already included.

Hmmm this would be a similar concept to how Twitter does it, it should work, only one way to find out! *Just merge it* lolol

I know the meta is injecting properly like the the titles, I would think in theory it should work, no way to really test it.

tijno:
i can put it live in my node later and test it with that

carsenk:
> i can put it live in my node later and test it with that

Good point I should just setup a test domain...

Let me know how it fairs! Thanks!

tijno:
Just to confirm discussion on discord with @carsenk 

The only way for this to work is:

1) server renders ograph meta tags for each page request
2) bitclout becomes a oEmbed provider and runs a dedicated endpoint for embed requests that returns rendered html for the embed.

The dynamic JS title changes are still useful I think for using multiple tabs in the browser, but I dont see a purpose for the dynamic meta tags.

maebeam:
Thanks for the PR! I can confirm that we don't currently do server-side rendering of pages which means open embed links will still just say "Welcome to BitClout." The dynamic page titles are great though!

src/app/admin/admin.component.html
```diff
@@ -1,4 +1,5 @@
 <div *ngIf="globalVars.nodeInfo != null">
+  {{ setTitle('Admin - BitClout') }}
```
maebeam, line 2: Please move all setTitle calls from html files to `ngOnInit`

src/app/creator-profile-page/creator-profile-details/creator-profile-details.component.ts
```diff
@@ -199,4 +202,22 @@ export class CreatorProfileDetailsComponent {
...
     );
   }
+
+  // Set Title function for dynamically setting the title
+  public setTitle(newTitle: string) {
+    this.titleService.setTitle(newTitle);
+  }
+
+  // Set Meta function for dynamically setting the meta tags
+  public setMeta(newContent: string) {
```
maebeam, line 212: Sorry, but could you remove the dynamic meta tags as well? They will need to be done differently when we move to server-side rendering.

maebeam (changes requested):

src/app/admin/admin.component.html
carsenk: Will do!

src/app/creator-profile-page/creator-profile-details/creator-profile-details.component.ts
carsenk: Yes I will just update this other PR: https://github.com/bitclout/frontend/pull/97 which has the titles, metatags were removed

carsenk:
Basically moved this to this PR (removed Metatags) and updated recommended changes for now

https://github.com/bitclout/frontend/pull/97

PR #107, opened 2021-06-08
Create Themable Themes (in CSS not TS)

JustifydWarrior:
* move themes out of inline styles and into (s)css
* simplify dropdown to use a container class rather than app state & ts to hold styles
* make editing & adding themes very simple: 
  * copypasta some existing theme & rename
  * add entry to select box with class name as value   
  
 live at [bitclout.pixelbypixelstudios.live](https://bitclout.pixelbypixelstudios.live)


teambrixmusic:
@JustifydWarrior This is really cool

maebeam:
This is great! If you'd like to update this PR I'd love to get it merged 💪🏼

JustifydWarrior:
Sorry, I've been neck deep implementing Twitch Drops for our game. This should be up to date again.

My employer pushed pause, but not stop, on running a public frontend for our node, which is why it's been pretty quiet on my end.

src/app/theme/themes/legends.scss
```diff
@@ -0,0 +1,23 @@
+// Ledgends Theme
+// By:
+.ledgends[app-theme] {
```
lazynina, line 3:
```suggestion
.legends[app-theme] {
```

src/app/theme/themes/legends.scss
```diff
@@ -0,0 +1,23 @@
+// Ledgends Theme
```
lazynina, line 1:
```suggestion
// Legends Theme
```

src/app/theme/themes/_index.scss
```diff
@@ -0,0 +1,6 @@
+@import 'light.scss';
+@import 'dark.scss';
+@import 'icydark.scss';
+@import 'legends.scss';
+@import 'cake.scss';
+@import 'greenish.scss';
```
lazynina, line 6: We need to copy over the cake and greenish themes into their own scss files.

lazynina (changes requested):

lazynina:
also need to add the `--button` property

src/app/theme/themes/_index.scss
JustifydWarrior: that was a big miss, just forgot to commit them.

src/app/theme/themes/legends.scss
JustifydWarrior: oops we're not looking for "ledge ends"

JustifydWarrior:
should be all set. nice catch on the buttons. they got added and caused a regression in the older scss.

lazynina:
The themes are working now. However, this PR causes the page to refresh when a new theme is selected whereas the current implementation does not reload the page. Is there a way to accomplish the theme change without a reload here?

JustifydWarrior:
yes there is.

JustifydWarrior:
@lazynina it's working as expected now. I tested side by side with the dot com site several times.

lazynina (approved):
looks good to me! @maebeam for second review

teambrixmusic:
Awesome! So exciting!

maebeam (approved):
Thank you so much for cleaning this up!

PR #108, opened 2021-06-08
Enable super admins to fetch admin logs on users.

bluepartyhat:


maebeam (approved):

src/app/backend-api.service.ts
```diff
@@ -1166,6 +1167,12 @@ export class BackendApiService {
     });
   }
 
+  AdminGetUserAdminData(endpoint: string, AdminPublicKey: string, UserPublicKeyBase58Check: string): Observable<any> {
+    return this.jwtPost(endpoint, BackendRoutes.RoutePathAdminGetUserAdminData, AdminPublicKey, {
+      UserPublicKeyBase58Check,
```
lazynina, line 1172: You need to include `AdminPublicKey` in the body otherwise this endpoint will fail when a node has >0 admin public keys

lazynina (changes requested):

lazynina (approved):

PR #110, opened 2021-06-09
Add Create Post Button to Left Bar

carsenk:
Adds a "Create Post" button to the left bar navigation. Basically extremely useful to be able to post easily.

### DIFF
--- a/src/app/left-bar/left-bar.component.html
+++ b/src/app/left-bar/left-bar.component.html
@@ -91,6 +91,15 @@
         <a class="cursor-pointer fs-15px text-grey5" [href]="getHelpMailToAttr()" target="_blank">Help</a>
       </div>
 
+      <a
+        [routerLink]="'/' + globalVars.RouteNames.CREATE_POST"
+        queryParamsHandling="merge"
+      >
+      <div class="w-80 justify-content-center d-flex btn-primary text-white" style="border-radius:10px; padding:10px; margin: 0 auto; margin-top: 20px;">
+        Create Post
+      </div>
+      </a>
+
       <div class="p-15px"></div>
 
       <left-bar-button


lazynina:
this doesn't stylistically match the left sidebar and I don't know that this is a necessity.  Desktop users can create posts from the bitclout.com/browse and redirecting them to a different page doesn't seem to be the most effective solution. 

If we were to add some ability to create a post from anywhere in the app, I would prefer that we open a modal so we don't take users out of their current place in the application. Additionally, I'd like to see this  option in desktop only as we already have the + sign to create a post on mobile (or we remove that option from mobile). 

@maebeam thoughts?

carsenk:
> this doesn't stylistically match the left sidebar and I don't know that this is a necessity. Desktop users can create posts from the bitclout.com/browse and redirecting them to a different page doesn't seem to be the most effective solution.
> 
> If we were to add some ability to create a post from anywhere in the app, I would prefer that we open a modal so we don't take users out of their current place in the application. Additionally, I'd like to see this option in desktop only as we already have the + sign to create a post on mobile (or we remove that option from mobile).
> 
> @maebeam thoughts?

+1 to the modal idea for creating new posts outside of /browse, it is just a pain on desktop to have to click home each time which is small text in navigation. Easier certainly to hit a big button and less confusing to new users. We could make the color fit more into the theme that it is selected too rather than a bright blue and have it popup a modal.

lazynina:
> > this doesn't stylistically match the left sidebar and I don't know that this is a necessity. Desktop users can create posts from the bitclout.com/browse and redirecting them to a different page doesn't seem to be the most effective solution.
> > If we were to add some ability to create a post from anywhere in the app, I would prefer that we open a modal so we don't take users out of their current place in the application. Additionally, I'd like to see this option in desktop only as we already have the + sign to create a post on mobile (or we remove that option from mobile).
> > @maebeam thoughts?
> 
> +1 to the modal idea for creating new posts outside of /browse, it is just a pain on desktop to have to click home each time which is small text in navigation. Easier certainly to hit a big button and less confusing to new users. We could make the color fit more into the theme that it is selected too rather than a bright blue and have it popup a modal.

I think having the color match the theme and having the button resized (it's bit large IMO) would help with this issue. I would still push for modal instead of directing to the "create post" page, which is a bit bare and unfriendly.

iPaulPro:
It's a feature that people really like in BitClout+. Using the modal is actually not ideal because it's very easy to accidentally dismiss it and lose your draft. 

People are constructing posts with a thousand or more characters and having a dedicated draft page is quite valuable. 

It should certainly be a secondary button, as far as style.

carsenk:
> It's a feature that people really like in BitClout+. Using the modal is actually not ideal because it's very easy to accidentally dismiss it and lose your draft.
> 
> People are constructing posts with a thousand or more characters and having a dedicated draft page is quite valuable.
> 
> It should certainly be a secondary button, as far as style.

Fixed up the styling good to go there, Yes I agree with you on that @iPaulPro to think about that more, accidentally dismissing the modal window would be such a pain if deep in a draft... So +1 to keeping it as separate page like it is

carsenk:
@lazynina @maebeam thoughts? Button now matches theme and is a bit more integrated looking

PR #113, opened 2021-06-09
switch to $clout

tijno:
With 3 days to go to deflation, i thought it would be good to switch to $clout instead of $BitClout.

20210609-fvhFOrEY

Replaced all occurrences apart from a few exceptions, like landing page header where I changed from `$BitClout` to `BitClout`.


lazynina (changes requested):
Can we change from $Clout to $CLOUT please? Also, resolve the merge conflicts.

tijno:
Ill resubmit this - too many changes since i did this, so will resubmit.

PR #117, opened 2021-06-10
Ln/get single profile picture

lazynina:


src/app/app.component.ts
```diff
@@ -321,6 +318,11 @@ export class AppComponent implements OnInit {
       this._updateBitCloutExchangeRate();
     }, 5 * 60 * 1000);
 
+    // Clear the avatar map every 5 minutes
```
maebeam, line 321: Can we just require a reload to get fresh avatars? I am worried this will spam the servers with requests every 5 mins

maebeam (approved):
incredible! been wanting this for so long

PR #118, opened 2021-06-10
Add Founder Reward Percentage to Profile Pages

carsenk:
Adds in the Founder Reward percentage into profile pages.

Preview
28c85d4f5d


lazynina (changes requested):
I think this is a good change - one small change requested

carsenk:
> I think this is a good change - one small change requested

Done 💎🚀

lazynina (approved):

PR #120, opened 2021-06-10
Updates to Wallet Page

carsenk:
This adds in a Total Balance at the top of the Wallet page along with the total count of creator coins held and clout balance information. Also added missing "CLOUT" tags, if we want to use a different ticker feel free to lmk @maebeam 

Preview:
walletupdatesfkdljfskldf

### DIFF
--- a/src/app/wallet/wallet.component.html
+++ b/src/app/wallet/wallet.component.html
@@ -5,6 +5,7 @@
   >
     <top-bar-mobile-navigation-control class="mr-15px d-lg-none d-inline-block"></top-bar-mobile-navigation-control>
     Wallet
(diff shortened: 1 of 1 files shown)


carsenk:
@maebeam @lazynina thoughts on this PR also? Should be good to go if wanted

lazynina (changes requested):
I like these additions to the wallet, but would like to use $CLOUT instead of just CLOUT.  We're trying to move to getting two reviews before merging, so will wait on @maebeam for a second review.

lazynina (approved):

maebeam (approved):

PR #130, opened 2021-06-12
copy-paste image into text area will upload image. add bombfetti

lazynina:


tijno:
not lazy at all :) brootle will be delighted.

AdityaChaudhary0005:
Hopefully I won't be opening explorer to choose image now haha!

Nina isn't lazy :)

PR #136, opened 2021-06-12
Fix profile avatars on notifications

carsenk:
@lazynina @tijno 

I noticed also profile avatars on profile pages will display both the default avatar and the users avatar over it if it is transparent, I can open another PR for a fix if needed?

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -1417,6 +1417,7 @@ $green-font-color: #19B028;
    height: 28px;
    width: 28px;
    border-radius: 3px;
+   background-size: auto 100%;
 }
 
 .notifications__action {


lazynina (approved):

lazynina:
> @lazynina @tijno
> 
> I noticed also profile avatars on profile pages will display both the default avatar and the users avatar over it if it is transparent, I can open another PR for a fix if needed?

Makes sense if the background is transparent, just need to make sure that the default profile photo is used correctly in components such as the notifications list.

carsenk:
> > @lazynina @tijno
> > I noticed also profile avatars on profile pages will display both the default avatar and the users avatar over it if it is transparent, I can open another PR for a fix if needed?
> 
> Makes sense if the background is transparent, just need to make sure that the default profile photo is used correctly in components such as the notifications list.

Maybe needs some more logic I would think within the setAvatar() to check if avatar is transparent or not to load in default...or maybe if set at all do not load in default

tijno:
would better solution be to let backend return default avatar, or 302 redir to default if none is set @lazynina

PR #151, opened 2021-06-14
fix styling of the posts-by-clout select to support dark and icy dark  modes, set background on option elements according to selected theme

lazynina:
image
image
Screenshot from 2021-06-13 18-26-47


carsenk:
Looks much better! 👍 💯

PR #166, opened 2021-06-14
if btc ticker fails to respond, set the usdPerBitcoinExchangeRate to the Protocol USD Per Bitcoin Exchange rate

lazynina:


src/app/app.component.ts
```diff
@@ -162,29 +162,36 @@ export class AppComponent implements OnInit {
...
-          this.ref.detectChanges();
+    this.httpClient
+      .get<any>("https://api.blockchain.com/ticker")
+      .subscribe(
+        (res: any) => {
+          if (res.USD != null && res.USD.last != null) {
+            this.globalVars.usdPerBitcoinExchangeRate = res.USD.last;
+          } else {
+            // We divide by 100 to convert from cents to dollars. usdPerBitcoinExchangeRate is expected to be in dollars.
+            this.globalVars.usdPerBitcoinExchangeRate = this.globalVars.ProtocolUSDCentsPerBitcoinExchangeRate / 100;
```
maebeam, line 173: Do we know that `ProtocolUSDCentsPerBitcoinExchangeRate` is populated when this call happens? E.g. does this call happen after `GetExchangeRate`?

lazynina: yup -- the only usage of this function is in the subscription to `_updateBitCloutExchangeRate` which calls `GetExchangeRate`

maebeam (approved):

PR #167, opened 2021-06-15
Queue up diamonds and then create transactions serially for all diamonds in queue every 5 seconds

lazynina:
This prevents users from encountering the `You're doing that a bit too quickly. Please wait a second or two and try again.` error when giving diamonds in rapid succession.


lazynina:
coming up with a more elegant solution

PR #169, opened 2021-06-15
Fix Add a comment/Add your quote

carsenk:
Fix Add a comment/Add your quote

These were opposite of what they should be 👍 @lazynina @maebeam


lazynina (approved):
:woman_facepalming: thanks!

PR #171, opened 2021-06-15
 Set target for SweetAlert and Diamond popover to parent container within app-root

lazynina:
set target for sweet alert modals and diamond popover to be a parent container within app-root so it has access to themes, add target in all calls to swal helper, remove diamond popover classes specific to themes


src/app/admin-page/admin-page.component.html
```diff
@@ -1,11 +1,11 @@
 <app-page>
   <admin
     *ngIf="globalVars.showAdminTools()"
-    class="h-100 center-col-width border-left border-right border-color-grey"
+    class="h-100 center-col-width"
```
lazynina, line 4: there was an extra wide border at the top of the admin page that was larger than the rest of the page.

src/app/not-found-page/not-found-page.component.html
```diff
@@ -1,30 +1,3 @@
...
-  <left-bar-mobile></left-bar-mobile>
-
-  <div style="flex-grow: 1">
-    <not-found></not-found>
-    <div class="global__bottom-bar-mobile-height">&nbsp;</div>
-  </div>
-
-  <bottom-bar-mobile class="global__bottom-bar-mobile"></bottom-bar-mobile>
-</div>
+<app-page>
```
lazynina, line 1: simplify the not-found page by using app-page

maebeam (approved):

PR #180, opened 2021-06-16
Ln/profile cleanup

lazynina:
moved coin price to the top and removed coin price in clout to reduce information overload based on feedback from @diamondhands0 

image

image


carsenk:
Missing clout price?

I like the new area of coin price, should it be in a box though? To separate further from bio?

maebeam (approved):

PR #196, opened 2021-06-19
add support for embedding soundcloud

lazynina:


maebeam (approved):
It'd be great if we could add some `invalidSoundCloudURLs` to the tests

lazynina:
> It'd be great if we could add some `invalidSoundCloudURLs` to the tests

added! good catch thanks!

PR #197, opened 2021-06-19
New Emoji Picker for Posts

carsenk:
This adds in a new Emoji icon next to embeds and images for posting, which brings up a nice selector to select an emoji, it can go off system themes only (light or dark) but we could expand this eventually...Seems to work great on mobile as well, this uses unicode emojis so they fit right into your current message.

This does have a new dependency required ` npm install @joeattardi/emoji-button` so you will need to do a fresh npm install of the frontend for this to work, otherwise should be good to go @lazynina @maebeam @diamondhands0 

Preview:
0JtjRoQYEL


src/styles.scss
```diff
@@ -641,7 +641,7 @@ $green-font-color: #19B028;
 .global__top-bar {
   position: fixed;
   background: var(--background);
-  z-index: 30;
+  // z-index: 30; Should not float above anything like emoji picker
```
lazynina, line 644: this breaks a lot of styling - try scrolling down on the "trends" page with this branch checked out. It looks something like this:
image

lazynina: instead let's add styling to `.emoji-picker__wrapper` and set it's z-index to be higher than z-index for `.global__top-bar`

src/app/feed/feed-create-post/feed-create-post.component.ts
```diff
@@ -71,6 +72,25 @@ export class FeedCreatePostComponent implements OnInit {
...
+    const trigger = event.currentTarget;
+    const picker = new EmojiButton({
+      theme: window.localStorage.getItem("theme") == "light" ? "light" : "dark",
+      position: "auto",
+      autoHide: false,
+      emojiSize: "1.3em",
+      emojisPerRow: 6,
+      rows: 3,
+      initialCategory: "smileys",
+      showPreview: false,
```
lazynina, line 85: let's set z-index here actually and just make it greater than the z-index of the global__top-bar

src/app/feed/feed-create-post/feed-create-post.component.ts
```diff
@@ -71,6 +72,25 @@ export class FeedCreatePostComponent implements OnInit {
     this._setRandomMovieQuote();
   }
 
+  emojiClick(event: any): void {
+    const trigger = event.currentTarget;
+    const picker = new EmojiButton({
+      theme: window.localStorage.getItem("theme") == "light" ? "light" : "dark",
```
lazynina, line 78: we should also set `rootElement` to some element in feed-create-post. for example, I set the id on the first div in `feed-create-post.component.html` and got that element by id and used that as the root element.  By doing this, we can pass all the theme variables to the emoji picker. 

Do this and then add styling to `.emoji-picker` to set background color, font colors, and borders.

lazynina: rootElement described here: https://emoji-button.js.org/docs/api/

lazynina (changes requested):
See feedback comments please.  Also, this may take a while as there are dependencies to be audited for vulnerabilities.

src/app/feed/feed-create-post/feed-create-post.component.ts
carsenk: > we should also set `rootElement` to some element in feed-create-post. for example, I set the id on the first div in `feed-create-post.component.html` and got that element by id and used that as the root element. By doing this, we can pass all the theme variables to the emoji picker.
> 
> Do this and then add styling to `.emoji-picker` to set background color, font colors, and borders.

I added rootElement in, I am not sure how you are getting the element ID to set as root, so I will let you do that, or we can do in future to extend the theming.

src/app/feed/feed-create-post/feed-create-post.component.html
```diff
@@ -1,7 +1,7 @@
 <!-- Be careful when modifying this.
      It's used by both the feed and that "create post page."-->
 
-<div [ngStyle]="{'background-color': !parentPost ? 'var(--formbg)' : ''}">
+<div [ngStyle]="{ 'background-color': !parentPost ? 'var(--formbg)' : '' }">
```
lazynina, line 4:
```suggestion
<div id="feed-post-input-container" [ngStyle]="{ 'background-color': !parentPost ? 'var(--formbg)' : '' }">
```

src/app/feed/feed-create-post/feed-create-post.component.ts
```diff
@@ -71,6 +72,27 @@ export class FeedCreatePostComponent implements OnInit {
     this._setRandomMovieQuote();
   }
 
+  emojiClick(event: any): void {
+    const trigger = event.currentTarget;
+    const picker = new EmojiButton({
+      rootElement: document.body,
```
lazynina, line 78:
```suggestion
      rootElement: document.getElementById("feed-post-input-container"),
```

src/styles.scss
```diff
@@ -2220,3 +2220,9 @@ input {
   background-color: var(--background) !important;
   color: var(--text);
 }
+
+.emoji-picker__search-icon {
+  top: calc(50% - 0.6em) !important;
+}
+
+
```
lazynina, line 2228:
```suggestion
.emoji-picker {
  background-color: var(--background) !important;
  color: var(--text);
  border-color: var(--border) !important;
}

.emoji-picker__search {
  border-color: var(--border) !important;
  color: var(--text);
}

button.emoji-picker__category-button {
  color: var(--grey) !important;
}

button.emoji-picker__category-button.active {
  color: var(--cblue) !important;
  border-bottom-color: var(--cblue);
}
```

lazynina:
@carsenk I suggested some changes that should apply the selected theme to the emoji picker.  This is an example of what it looks like on legends theme with the suggested changes

image

carsenk:
> @carsenk I suggested some changes that should apply the selected theme to the emoji picker. This is an example of what it looks like on legends theme with the suggested changes
> 
> image

Pushed changes up, works across themes! 🚀🤘

lazynina (approved):
@maebeam for 2nd review.  I looked through the dependencies of the emoji-button lib and they appear to be well-maintained and safe from my review.

maebeam:
While this PR looks great, our `package-lock.json` file is currently frozen until we implement LavaMoat to protect against dependency tree compromises.

carsenk:
> While this PR looks great, our `package-lock.json` file is currently frozen until we implement LavaMoat to protect against dependency tree compromises.

How do you want us to proceed? package-lock.json should be in gitignore?

PR #199, opened 2021-06-21
remove items from caddyfile

lazynina:


Caddyfile
```diff
@@ -67,10 +67,6 @@ header @html Content-Security-Policy "
     https://www.tiktok.com
     https://giphy.com
     https://open.spotify.com
-    https://player.twitch.tv
-    https://clips.twitch.tv
     https://w.soundcloud.com
     pay.testwyre.com
-    pay.sendwyre.com;
-  frame-ancestor 'self'
```
maebeam, line 75: Wait don't we want to keep frame-ancestor?

lazynina: It wasn't there before - so I removed it. I was only trying to add SoundCloud. Is there a reason we do want frame-ancestors?

maebeam: I have a note to add it and I thought you were doing it for me :)

lazynina: hahaha okay I'll add it back. I gotchu :heart:

PR #207, opened 2021-06-25
Fix top shadow on search bar on iOS

michailbrynard:
@lazynina This fixes the thicker top border/ shadow issue on the search bar on iOS. 

The previous fix https://github.com/bitclout/frontend/pull/137 does not seem to work.

Before:
Screen Shot 2021-06-25 at 12 05 13 PM

After:
Screen Shot 2021-06-25 at 12 04 35 PM


maebeam:
Thanks!

PR #209, opened 2021-06-28
show user to whom you are sending clout on the send-bitclout page

lazynina:
Use search-bar component to get username and avatar when searching for user on Send BitClout page.

image

image

Updated search bar component to search for single profile based on public key if search text looks like a public key.

Supports public keys with a profile:
image

### DIFF
--- a/src/app/search-bar/search-bar.component.ts
+++ b/src/app/search-bar/search-bar.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnInit, Renderer2, ViewChild, ElementRef, Input, Output, EventEmitter } from "@angular/core";
 import { GlobalVarsService } from "../global-vars.service";
 import { ActivatedRoute, Router } from "@angular/router";
-import { BackendApiService } from "../backend-api.service";
+import { BackendApiService, ProfileEntryResponse } from "../backend-api.service";
 import * as _ from "lodash";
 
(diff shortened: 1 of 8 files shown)


src/app/search-bar/search-bar.component.ts
```diff
@@ -45,7 +54,44 @@ export class SearchBarComponent {
       readerPubKey = this.globalVars.loggedInUser.PublicKeyBase58Check;
     }
 
-    this.backendApi
+    // If we are searching for a public key, call get single profile with the public key.
+    if (
```
maebeam, line 58: Do we do these checks anywhere else in the repo? If so, worth making a helper?

maebeam (approved):
Looks great!

maebeam:
How does this look for blacklisted users?

src/app/search-bar/search-bar.component.ts
lazynina: we have a function called `isMaybePublicKey` :joy: I'll update it to include length checks and use it here.

lazynina:
> How does this look for blacklisted users?

line 66 in `search-bar.component.ts`, if a user is blacklisted is returned by the get-single-profile call, we don't display that user.  You will not be able to send to blacklisted keys.

The call to `/get-profiles` already filters out blacklisted public keys for us.

PR #212, opened 2021-06-28
fix waiting for multiple transactions to broadcast

lazynina:


src/app/app.component.ts
```diff
@@ -228,9 +228,18 @@ export class AppComponent implements OnInit {
     // If we have a transaction to wait for, we do a GetTxn call for a maximum of 10s (250ms * 40).
     // There is a success and error callback so that the caller gets feedback on the polling.
     if (waitTxn !== "") {
-      this._repeatForX(
```
maebeam, line 231: Can we deelte the `_repeatForX` function and `repeatForXInterval` var?

lazynina: done!

PR #230, opened 2021-07-06
cake-theme.ts

teambrixmusic:
Hey there! Don't expect this to actually merge and not certain I remember what I'm doing but Tijn inspired me to do a PR and I loved Carsenk themes so (I'm not completely sure what this looks like yet) but this is an alternative lightTheme in festive childlike pastels colors. The luminosity of the original theme colors are relatively the same so the contrast shouldn't change. Simply, the whites and greys have become slightly tinted pinks and blues and the darker colors have taken on shades of purple and blue. It's a bit whacky, and intended on looking like a silly myspace-ish birthday cake throwback— perhaps for fun, to use on your Birthday. Or a teen might like it.

In all seriousness, when my node is back up and I figure out Git on Windows, I'll have more to add. 

Won't take a rejection personally, was just for fun and to learn again. <3


lazynina:
image
had to make a few modifications to get it working, but here's what it looks like if you want to see! 

You'll need to modify `app.module` with the following lines:
around line 128
```
import { cakeTheme } from "./theme/cake-theme";
```

and modify around line 250
```
      themes: [lightTheme, darkTheme, icydarkTheme, legendsTheme, cakeTheme],
```

lazynina (changes requested):
see my feedback above

teambrixmusic:
> see my feedback above

Ah! I wasn't sure which classes went to which. Thanks! On it. Will give it a shot. :) BRB

teambrixmusic:
I think I did it right this time lol. Let me know.

cake

teambrixmusic:
Your screenshots look so much better than mine @lazynina 
Is the logo filter too pink? It matches the hearts (hot pink)
@carsenk Wish we could break out more theme classes i.e. .feed-post__blue-border & .tab-underline-active are very visible selection markers. Even if it could match link color by default.
Just realized how fun this would look with the confetti animation lol!

src/app/theme/cake-theme.ts
```diff
@@ -0,0 +1,27 @@
...
+    "--secondary": "#f3e1e1",
+    "--secalt": "#fae1dd",
+    "--textalt": "#a97aa2",
+    "--norm": "#47205c",
+    "--formbg": "#fff5f3",
+    "--link": "#009aff",
+    "--hover": "#008cad",
+    "--border": "#fae1dd",
+    "--mborder": "#fae1dd",
+    "--filter": "blur(0.6px) grayscale(0.5) sepia(1) hue-rotate(227deg) saturate(9.2) opacity(0.78)",
```
lazynina, line 19:
```suggestion
    "--filter": "grayscale(0.5) sepia(1) hue-rotate(227deg) saturate(9.2) opacity(0.78)",
```
the blur seems unnecessary and just makes the logo look strange.

lazynina (approved):

PR #243, opened 2021-07-07
Add Option for Change Button Color as Per Theme

harpreet9629:
Screenshot_1
Screenshot_2
Screenshot_3
Screenshot_4

It make looks nice :)


teambrixmusic:
YES! THANK YOU!!!

teambrixmusic:
@harpreet9629 This is awesome and love the change with the select BAR (under tabs) and button possibly matching everything. Love the added --button also. However, I think you may have accidentally modified cake-theme.ts "Link" color to dark purple which is too similar to the regular text. The changes to Cake effectively remove all blue, which was custom to the theme (and not just because the button option wasn't available.) Love that future themes would be able to implement this option <3

If this merges I'll probably end up pulling again to readjust the cake theme per the original. Any way you could just revert the link 009aff , hover 008cad and new button 009aff back to blue since it isn't merged yet?

Love the gray theme too!

harpreet9629:
Done i revert theme colors back!

teambrixmusic:
> Done i revert theme colors back!
@harpreet9629 
I love what you've done, by the way! That opens things up for future theming options which is awesome. I too am frustrated by the limited number of class options. I thought it would also be simple someday to implement other google fonts since the themes primarily use Roboto. Also might be cool to have avatars someday square | rounded | round but I haven't looked at everything.

Do you have a screenshot of gray?

harpreet9629:
Screenshot_5
It's Green not gray, i will change theme name.

src/styles.scss
```diff
@@ -139,15 +139,15 @@ $border-grey: #E5E5E5;
 $staking-grey: #F0F0F0;
 $light-grey-for-divider: #F8F8F8;
 $light-grey-for-divider-2: #FAFAFA;
-$buy-button-blue: #0058F7;
```
lazynina, line 142: let's use this hex value as the button color for light, dark, and icy dark themes.

src/styles.scss
```diff
@@ -1999,10 +1999,20 @@ input {
...
+  background-color: $buy-button-blue;
+  border-color: $buy-button-blue;
   color: #fff;
 }
+.btn-primary.disabled, .btn-primary:disabled {
+	color: #fff;
+	background-color: $buy-button-blue;
+	border-color: $buy-button-blue;
+}
+.btn-primary:hover {
```
lazynina, line 2011: for some reason, when you hover over a disabled button, it changes color, which it shouldn't do when disabled.

lazynina (changes requested):

src/app/theme/gray-theme.ts
```diff
@@ -0,0 +1,28 @@
+import { Theme } from "./symbols";
```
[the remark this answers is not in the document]
harpreet9629: Ok i will do remove

src/styles.scss
harpreet9629: Ok got it cause hover is after disabled class

src/styles.scss
harpreet9629: Ok sure

harpreet9629:
@lazynina its fixed 👍

lazynina (approved):

Issue #221, opened 2021-07-04
Not reloading feed


harpreet9629:
When I switch to another account it won't refresh/reload feed, I saw my previous account clouts. (I liked, reclouted, diamond) all show previous account details.

AdityaChaudhary0005:
yeah, i have experienced this sometimes.
the only solution is to reload the page.

harpreet9629:
But i can be fix as same when change name and page it reload feed also on switch account

tijno:
If you know fix - best is to submit a PR with the fix. Its more likely to get fixed then.

PR #244, opened 2021-07-07
When switching user stay on same page and reload

cat-ico:
When an account is switched the app always navigates to the browse page. This a bit user-unfriendly.
Also, the data is not reloaded, so when user A likes a post on a feed and switches to user B the like button stays toggled on in the UI.

This PR changes the behavior so that:
* The user stays on the same page when the account is switched
* The route is reloaded so that the data is updated for the current user

This also fixes issue #221

### DIFF
--- a/src/app/change-account-selector/change-account-selector.component.ts
+++ b/src/app/change-account-selector/change-account-selector.component.ts
@@ -43,8 +43,8 @@ export class ChangeAccountSelectorComponent {
     this.globalVars.messageResponse = null;
     this.globalVars.SetupMessages();
 
-    this.router.onSameUrlNavigation = 'reload'
-    this.router.navigateByUrl(this.router.url);
+    const currentUrl = this.router.url
+    this.router.navigate(['/']).then(() => { this.router.navigateByUrl(currentUrl); })
 
     this.globalVars.isLeftBarMobileOpen = false;
   }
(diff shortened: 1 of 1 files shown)


lazynina (changes requested):
These lines alone don't refresh all the data based on the newly selected user.  For example, if I'm on the notifications page and I switch users, the notifications don't reload. 

The below line, if added to the constructor in this component, will reload the page and all data based on the new user.
```
this.router.routeReuseStrategy.shouldReuseRoute = () => false;
```

maebeam:
Do we want to set `shouldReuseRoute` to `false` for all routes? If yes, let's see if we can do that once globally. If not, perhaps the right solution might be something like

```
this.router.navigate(['/']).then(() => { this.router.navigate([newRouterLink ]); })
```

cat-ico:
@lazynina Strange, when I go to post detail and then to notifications, the reloading works as expected. But when I start at the notifications page it doesn't work.

I updated the fix as per @maebeam's suggestion.

Thank you both!

maebeam (approved):
Motto of the year is angular sucks. Looks good to me!

lazynina (approved):

brootle:
> Motto of the year is angular sucks. Looks good to me!
you want to deploy ReactJS version of Bitclout client web app? the number of PRs would explode, seriously.

maebeam:
@brootle I don't want to port the app to React but someone should!

PR #258, opened 2021-07-10
Allow blocking/viewing of blacklisted profiles

derishaviar:
This allows users to block a blacklisted user from spamming their notifications, as well as checking a blacklisted user's coin holders.

This does not change the search or feed's blacklisting functionality.

### DIFF
--- a/src/app/creator-profile-page/creator-profile-details/creator-profile-details.component.ts
+++ b/src/app/creator-profile-page/creator-profile-details/creator-profile-details.component.ts
@@ -166,7 +166,7 @@ export class CreatorProfileDetailsComponent {
     this.loading = true;
     this.backendApi.GetSingleProfile(this.globalVars.localNode, "", this.userName).subscribe(
       (res) => {
-        if (!res || res.IsBlacklisted) {
+        if (!res) {
           this.loading = false;
           this.router.navigateByUrl("/" + this.appData.RouteNames.NOT_FOUND, { skipLocationChange: true });
           return;


tijno:
Well done on first commit. You got past the hardest part of contributing - setting up and making your first commit.

Im delighted by that!

And i welcome ability to see a blocked users profile - as i often used my own node for that.

Couple of things though

- i think title can be clarified as this change is really about making profiles for blacklisted users viewable

- i believe blacklisting is used to hide user profiles that requested to be deleted - this change would unhide them which i dont think is desirable

- this is medium level change because of above and probably would need to be discussed following new cips process - note this was only launched yesterday so you may not have seen it: https://github.com/bitclout/cips

Hopefully the core devs will shes more light on this.

derishaviar:
> Well done on first commit. You got past the hardest part of contributing - setting up and making your first commit.
> 
> Im delighted by that!
> 
> And i welcome ability to see a blocked users profile - as i often used my own node for that.
> 
> Couple of things though
> 
> * i think title can be clarified as this change is really about making profiles for blacklisted users viewable
> * i believe blacklisting is used to hide user profiles that requested to be deleted - this change would unhide them which i dont think is desirable
> * this is medium level change because of above and probably would need to be discussed following new cips process - note this was only launched yesterday so you may not have seen it: https://github.com/bitclout/cips
> 
> Hopefully the core devs will shes more light on this.

Hi Tijn, yes thank you for encouraging me to learn something new - this was definitely out of my comfort zone!  Also thanks for the feedback, I will try to change the title. I'm also not aware of the cips or what they actually are, I'll check it out later :)

AdityaChaudhary0005:
great change @derishaviar, I recall a user who was blacklisted on main node but he was still spamming into my notifications and I was not able to block him.

derishaviar:
> great change @derishaviar, I recall a user who was blacklisted on main node but he was still spamming into my notifications and I was not able to block him.

Ah yes same, spamming notifications is really annoying and we should be able to not receive those notifications from who we manually select :)

lazynina:
@derishaviar - thanks for contributing! I understand the concern about blocking blacklisted profiles, but the purpose of blacklisting is to hide these profiles from the node which blacklisted the profile.  This change would break that constraint. Let's think of alternative ways we can block these profiles -- or better yet, make an update to the backend / core repos to remove blacklisted profiles from notifications (and anywhere else they aren't removed). I'm going to close this one out.

derishaviar:
> @derishaviar - thanks for contributing! I understand the concern about blocking blacklisted profiles, but the purpose of blacklisting is to hide these profiles from the node which blacklisted the profile. This change would break that constraint. Let's think of alternative ways we can block these profiles -- or better yet, make an update to the backend / core repos to remove blacklisted profiles from notifications (and anywhere else they aren't removed). I'm going to close this one out.

@lazynina Thank you for your reply and explanation. I will try to contribute better next time :)

PR #260, opened 2021-07-10
reclout button now shows sum of reclout and QuoteReclouts 

AdityaChaudhary0005:
The reclout button used to show reclout count only but just like any other platforms, reclout button should show sum of both reclouts and quote reclouts on a post.

preview (see the difference of count on reclout button ):
Pr3

### DIFF
--- a/src/app/feed/feed-post-icon-row/feed-post-icon-row.component.html
+++ b/src/app/feed/feed-post-icon-row/feed-post-icon-row.component.html
@@ -64,7 +64,7 @@
         </a>
       </ng-template>
     </div>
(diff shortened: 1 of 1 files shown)


harpreet9629:
Amazing

maebeam (approved):
Great work!

AdityaChaudhary0005:
Thanks @maebeam !

By the way, I have one question regarding the deployment part.

I don't think there is an auto-deployment for BitClout.com so I am curious to know how much time does it takes to re-deploy the changes which are merged in main?

maebeam:
There is no auto deployment. It usually takes less than 24 hours but can take up to a few days.

lazynina (approved):

PR #261, opened 2021-07-10
search for username even if there is @ in front or spaces around user…

brootle:
Simple update to handle situations when users try to put usernames including '@' and/or extra spaces in front and at the end in Bitclout search bar like for example  

'@SearchClout' 
'@SearchClout   '
'  @SearchClout   '
'   SearchClout'
'   SearchClout   '  

image

### DIFF
--- a/src/app/search-bar/search-bar.component.ts
+++ b/src/app/search-bar/search-bar.component.ts
@@ -93,7 +93,7 @@ export class SearchBarComponent implements OnInit {
         this.globalVars.localNode,
         "" /*PublicKeyBase58Check*/,
         "" /*Username*/,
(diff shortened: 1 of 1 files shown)


Shubs21:
subtle but useful!

gg @brootle

brootle:
> subtle but useful!
> 
> gg @brootle

thanks, I have some epic update in my mind, but need to learn more about Bitclout app structure, so why not fix small things too

harpreet9629:
Great PR @brootle

maebeam (approved):

lazynina (approved):
great work as always @brootle

PR #263, opened 2021-07-10
New Style of Icons in Notifications

harpreet9629:
Here's New Style for Icons in Notifications. I hope everyone likes this. :)

Screenshot_2

### DIFF
--- a/src/app/notifications-page/notifications-list/notifications-list.component.html
+++ b/src/app/notifications-page/notifications-list/notifications-list.component.html
@@ -13,20 +13,22 @@
 </div>
 
 <div #uiScroll *uiScroll="let item of datasource; let index = index" class="border-bottom border-color-grey">
-  <div class="p-10px cursor-pointer" *ngIf="item.action" [routerLink]="item.link">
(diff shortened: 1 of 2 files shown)


AdityaChaudhary0005:
I like it but i think previous icons were better since they emphasised more on type of notifications (i.e, likes/diamonds)
I am interested in knowing what others think about it.

teambrixmusic:
Tough call. I wonder how visible the icons would be on mobile at 100%
Esthetically, I think **this is far more attractive and modern**.
Functionally, I think it may render the icons too small on a phone.  :(

harpreet9629:
> Tough call. I wonder how visible the icons would be on mobile at 100%
> Esthetically, I think **this is far more attractive and modern**.
> Functionally, I think it may render the icons too small on a phone.  :( 
> 

I agree, with you but i have 1 more option to make it better,

And in this on mobile i can make bigger size, so it won't too small or hidden

AdityaChaudhary0005:
But what's wrong with the already existing notification icon?

I think it's unique and better in its own way :)

harpreet9629:
> But what's wrong with the already existing notification icon?
> 
> I think it's unique and better in its own way :) 

There's nothing wrong with existing ui, but its suggestion, if like it will will be approve otherwise it will close.

PR #274, opened 2021-07-12
Posts can't have JUST whitespaces

AdityaChaudhary0005:
BitClout.com allows users to do posts which have nothing but "JUST" white-spaces.
For example:
Screenshot from 2021-07-12 15-31-57

This PR solves this problem. Now users can't post with just blank-spaces. This applies to comments/quotes as well


tijno:
would this not stop you from posting a post with just an image - which i think is currently allowed?

AdityaChaudhary0005:
it's ok to post with just an image :) infact twitter has this feature too!

sometimes memers/artists would like to post an image without saying anything..

though i would like to listen the opinion of @lazynina or @maebeam

AdityaChaudhary0005:
oops, i misunderstood what you said Tijn.
yeah, you are right. i will change this ASAP

AdityaChaudhary0005:
I need to close this PR right now since i am not able to upload image from external node to test this. I will submit the PR later with successful testing.

PR #276, opened 2021-07-12
Update text that appears in Google results for "bitclout" 

enmyj:
The current text says what Bitclout isn't but not what it _is_

image


AdityaChaudhary0005:
I couldn't understand what "Own a piece by buying the BitClout coin $CLOUT" means.
though i would like to see this changed to some better and unique phrase.

tijno:
What about the one in the description meta tag ? 

https://github.com/bitclout/frontend/blob/07db3e8cc022d9918c4f1567e76ef704a1661635/src/index.html#L7

AdityaChaudhary0005:
awesome! short and strait-forward description is better in my opinion

teambrixmusic:
Agree with using the meta.

Issue #259, opened 2021-07-10
Implement middle click to open new tab on Bitclout.com


balajis:
Problem: right now you can't easily open multiple tabs on the Bitclout.com web interface by middle clicking the timestamp.

Solution: enable this behavior, just like Twitter.

tijno:
Whats the macOs equivalent ? right or cmd click i think.

i would love this !!

AdityaChaudhary0005:
As far I understand this problem, Balajis want the timeStamp as a link button so that one can open the link in new tab
![Screenshot from 2021-07-10 14-27-16](https://user-images.githubusercontent.com/55331140/125157883-1ab97e80-e18b-11eb-828b-506214a8b830.png)
right now there is no way to open post in new tab

chafreaky:
The behavior is mixed when for example clicking on a url in a post. The click will fire 2 events:
- one that will open the clicked link in a new tab
- another that will also open the current clicked post in the current page

I think this can be solved by `e.preventDefault()` on link clicks.

maebeam:
Fixing this behavior would be a massive win

AdityaChaudhary0005:
so @chafreaky you doing PR ? I tried myself but couldn't find the best possible solution.

chafreaky:
Angular is not my strong suit but I can give it a try.

jagath-ft:
I found the solution and it is working, the only issue that i notice here is "Page Title" is not getting updated in new tab, it is showing default title as "Welcome to BitClout" instead of eg: "balajis on Bitclout".
Working on to update page title dynamically on the fly, will keep you guys updated..

![image](https://user-images.githubusercontent.com/58214552/125430885-9fa064fd-741f-43ce-948c-145223658b68.png)

jagathks:
I found the solution and it is working, the only issue that i notice here is "Page Title" is not getting updated in new tab, it is showing default title as "Welcome to BitClout" instead of eg: "balajis on Bitclout".
Working on to update page title dynamically on the fly, will keep you guys updated..

![image](https://user-images.githubusercontent.com/17216156/125435901-100884bb-ac7d-4f98-b37d-a4af24a67590.png)

AdityaChaudhary0005:
Great work @jagathks !

What's your BitClout handle? Time to follow another gem!

maebeam:
Timestamp middle click has been fixed in #278 but it would be great to allow command + clicking (or windows equivalent) anywhere on the post.

[...]

PR #277, opened 2021-07-13
MatTooltip for Datetime on Feed Icon Row

jagathks:
As of now Feed shows how much old the feed is (in hours,min,days) but not showing the exact date time when the feed is posted.
I felt it should show datetime when we hover, so i added a MatTooltip.
preview (see the difference):
mat-datetime-tooltip

### DIFF
--- a/src/app/feed/feed-post-dropdown/feed-post-dropdown.component.html
+++ b/src/app/feed/feed-post-dropdown/feed-post-dropdown.component.html
@@ -1,6 +1,7 @@
 <div class="d-flex" dropdown (click)="$event.stopPropagation()" container="body" style="align-items: center">
   <div class="d-inline-block d-lg-none fs-12px fc-muted pr-5px">
-    {{ globalVars.convertTstampToDaysOrHours(postContent.TimestampNanos) }}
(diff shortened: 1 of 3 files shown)


AdityaChaudhary0005:
i like it! we were also wondering to make the timestamp a link to the post so that it can be opened in new tab!
https://github.com/bitclout/frontend/issues/259

jagath-ft:
Good idea, will work on it #259 .

jagathks:
Good idea, will work on it #259 .

lazynina (changes requested):
Please revert all whitespace / alignment changes - it's really hard to see what has actually changed here.

jagathks:
code updated, removed white space and alignment

lazynina (changes requested):
I like this - looks great on desktop. Can we add it to the timestamp that is in `feed-post-dropdown.component.html` so that this is available on mobile as well?

jagathks:
Changes are made in `feed-post-dropdown.component.html`

tijno:
great work @jagathks - i cant remember how often i went into webdev tools to find the exact post time for a post. This would solve that greatly.

jagathks:
> great work @jagathks - i cant remember how often i went into webdev tools to find the exact post time for a post. This would solve that greatly.

Exactly that is the problem i faced on my first day at bitclout (which is yesterday :) ) 
Glad you liked it...

lazynina (approved):
:+1:

PR #278, opened 2021-07-13
Issue #259 timestamp middle click

jagathks:
This pull request refers to fix for #259.

Now clicking on Timestamp with middle mouse button will open the post in new tab.

### DIFF
--- a/src/app/post-thread-page/post-thread/post-thread.component.ts
+++ b/src/app/post-thread-page/post-thread/post-thread.component.ts
@@ -59,7 +59,6 @@ export class PostThreadComponent {
     return new Datasource<IAdapter<any>>({
       get: (index, count, success) => {
         const comments = this.currentPost.Comments || [];
-        this.titleService.setTitle(this.currentPost.ProfileEntryResponse.Username + " on BitClout");
         if (!comments || (this.scrollingDisabled && index > comments.length)) {
           success([]);
           return;
--- a/src/app/post-thread-page/post-thread/post-thread.component.ts
+++ b/src/app/post-thread-page/post-thread/post-thread.component.ts
@@ -60,6 +60,7 @@ export class PostThreadComponent {
     return new Datasource<IAdapter<any>>({
       get: (index, count, success) => {
         const comments = this.currentPost.Comments || [];
+        this.titleService.setTitle(this.currentPost.ProfileEntryResponse.Username + " on BitClout");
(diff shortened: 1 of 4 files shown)


lazynina (changes requested):
please revert whitespace / alignment changes  so I can review more effectively.

jagathks:
changes made..

src/app/post-thread-page/post-thread/post-thread.component.ts
```diff
@@ -60,6 +60,7 @@ export class PostThreadComponent {
     return new Datasource<IAdapter<any>>({
       get: (index, count, success) => {
         const comments = this.currentPost.Comments || [];
+        this.titleService.setTitle(this.currentPost.ProfileEntryResponse.Username + " on BitClout");
```
lazynina, line 63: Does this need to be here? I see that we set the title in `ngOnInit`

jagathks: Yes it required to be here otherwise page title wont get updated, it will show the default one as "`Welcome to Bitclout`" instead of "`username at Bitclout`".
I think okay to remove this line from `ngOnInit` which is redundant in this case.

lazynina: I see - okay this should live in the function `refreshPosts` instead - This `get` is called frequently by the infinite scroller, to get the next page of data, whereas refreshPosts is only called once per post.

lazynina (changes requested):
see comment about moving titleService call to refreshPosts

jagathks (commented):
Ooppss.. I noticed see that, moved to refreshPosts() method.

src/app/post-thread-page/post-thread/post-thread.component.ts
```diff
@@ -60,6 +59,7 @@ export class PostThreadComponent {
...
```
lazynina, line 62:
```suggestion
```

lazynina (changes requested):
Remove title service from the datasource please. Will approve after that change

jagathks:
done

lazynina (approved):
approved - @maebeam for 2nd review / approval

maebeam (approved):

maebeam:
Could this same logic be used to make command + click work anywhere on the post? Would be great to be able to open a post in a new tab.

PR #284, opened 2021-07-13
Fix touch event listener and styling on change account selector

lazynina:


harpreet9629:
It make issue, Now can't able to open switcher

src/app/change-account-selector/change-account-selector.component.html
```diff
@@ -1,29 +1,28 @@
...
-    <div
-      *ngIf="!globalVars.loggedInUser"
-      class="font-italic change-account-selector__ellipsis-restriction cursor-pointer"
-    >
-      Logged out
+<div #changeAccountSelectorRoot  id="changeAccountSelectorRoot">
+  <div
+    class="fs-14px fc-default px-5px cursor-pointer change-account-selector br-4px change-account-selector__hover"
+    [ngClass]="{ 'change-account-selector__shadow': selectorOpen }"
+  >
```
AeonSw4n, line 5: Moving `#changeAccountSelectorRoot id="changeAccountSelectorRoot"` to the outer div causes #285. If you move it back into one div it works again.

src/styles.scss
```diff
@@ -873,10 +873,10 @@ $green-font-color: #19B028;
...
 	background-color: var(--secondary);
 	padding: 0px 0px 10px;
-	margin-top: 15px;
+	margin-top: 10px;
 	width: 100%;
-	margin-left: -4px;
-	z-index: 11;
+	margin-left: -10px;
+	z-index: 99;
 	border: 1px solid var(--border);
```
AeonSw4n, line 880: Also, `margin-left: -10px;` causes this behavior for me both on chrome and firefox: https://imgur.com/a/8TWyFie
If you change it to `margin-left: 0px;` and remove `border` prop it looks like this: https://imgur.com/a/VmkN0la
What do you think?

lazynina: I think #286 should resolve the problem. We should have been using a drop down directive long ago. I've tested that PR on a number of browsers, does it solve the issue you're having?

AeonSw4n: Yup, it works now. The only thing is, I feel like the dropdown became narrower. This causes the "blue dot" and the last 's' in "Send Creator Coins" to extend horizontally beyond the dropdown. Check it here: https://imgur.com/a/oUEYZKL
Maybe it could be a couple px wider?

harpreet9629: > I think #286 should resolve the problem. We should have been using a drop down directive long ago. I've tested that PR on a number of browsers, does it solve the issue you're having?  

Everything is fine as per my PR just have to change z-index:11 to z-index:99

PR #291, opened 2021-07-14
Fix redirecting when switching accounts

### LINKED ISSUE #289: I am redirecting tabs continue when switch accounts
First of all, Thank You @lazynina  and team Now Account Switcher working fine.

https://github.com/bitclout/frontend/pull/244

I think This PR making issue when I switch accounts, it continues redirecting tabs on the same page.

https://bitclout.com/browse?feedTab=Following to https://bitclout.com/browse?feedTab=Global
and Global tab to Following

Please look at this, it freezes the whole page and can't able to do any action.

@maebeam  @cat-ico

cat-ico:
Preserve selected tab when switching accounts on the feed page

Fixes #289

### DIFF
--- a/src/app/change-account-selector/change-account-selector.component.ts
+++ b/src/app/change-account-selector/change-account-selector.component.ts
@@ -43,7 +43,7 @@ export class ChangeAccountSelectorComponent {
     this.globalVars.SetupMessages();
 
     const currentUrl = this.router.url;
-    this.router.navigate(["/"]).then(() => {
+    this.router.navigate(["/" + this.globalVars.RouteNames.BROWSE]).then(() => {
       this.router.navigateByUrl(currentUrl);
     });
 


maebeam:
Could you help explain how this fixes the issue? I don't quite understand it at first glance

cat-ico:
@maebeam 

The issue is that the first navigation to "/" triggers a redirect on the landing page [here](https://github.com/bitclout/frontend/blob/main/src/app/landing-page/landing-page.component.ts#L112). As a result there are two parallel redirects happening at the same time and when there is a difference in query params for feedTab, it somehow results in some kind of redirection loop described in #289. I can't reproduce it every time.

By changing the first navigation to "/browse", the page reload will still work for all other routes, while preventing the reload while on feed. 

Feed is not currently reloading anyway (at least not completely), so I think it would be better to address the feed reloading in a separate PR.

Lmk, if I'm missing something obvious. (Sorry, it's my first time with angular.)

maebeam (approved):
Makes perfect sense. Thanks for explaining. I think this should do the trick.

PR #292, opened 2021-07-14
No more annoying lower case mentions and cashtags

AdityaChaudhary0005:
Mentions and Cashtags were in lower case before and to be honest it was annoying.
Idea by @HPaulson 

Before:
Screenshot from 2021-07-14 18-19-14

After:
Screenshot from 2021-07-14 18-19-22

PS. it works for $CashTags as well

### DIFF
--- a/src/lib/pipes/sanitize-and-auto-link-pipe.ts
+++ b/src/lib/pipes/sanitize-and-auto-link-pipe.ts
@@ -32,10 +32,6 @@ export class SanitizeAndAutoLinkPipe implements PipeTransform {
 
     // Only link usernames and cashtags for now (not hashtags etc)
     entities = _.filter(entities, (entity) => entity.screenName || entity.cashtag);
-    entities = _.each(entities, (entity) => {
-      entity.screenName = entity.screenName?.toLowerCase();
-      entity.cashtag = entity.cashtag?.toLowerCase();
-    });
 
     const textWithMentionLinks = twitter.autoLinkEntities(text, entities, {
       usernameUrlBase: `/${RouteNames.USER_PREFIX}/`,


harpreet9629:
> 
> 
> Does this use the users set case for their username, or does it just use the posters?
> 
> I think it's the latter.
> 
> So it would allow me to post `@ITSATiDyA` and it would still show up as such.
> 
> Not a massive issue - removing the lower casing definitely better. Especially when we have prepop mentions: [bitclout/cips#35](https://github.com/bitclout/cips/discussions/35)

I agree with @tijn

AdityaChaudhary0005:
It is dependent on the poster now. and i think it's fine. i just crosschecked it with other platforms like twitter and they work on the same principle :)

tijn:
> > Does this use the users set case for their username, or does it just use the posters?
> > I think it's the latter.
> > So it would allow me to post `@ITSATiDyA` and it would still show up as such.
> > Not a massive issue - removing the lower casing definitely better. Especially when we have prepop mentions: [bitclout/cips#35](https://github.com/bitclout/cips/discussions/35)
> 
> I agree with @tijn

@harpreet9629 Could you guys take care to name the right Tijn, please? The one with an extra o: @tijno, is the one you need. I understand that typos and small mistakes exists; it's fine, no bad feelings, but for me all these messages are basically spam in my inbox and I have no means to stop it other than hitting the unsubscribe button on each and every pull-request and issue where I'm mentioned by mistake.

harpreet9629:
> > > Does this use the users set case for their username, or does it just use the posters?
> > > I think it's the latter.
> > > So it would allow me to post `@ITSATiDyA` and it would still show up as such.
> > > Not a massive issue - removing the lower casing definitely better. Especially when we have prepop mentions: [bitclout/cips#35](https://github.com/bitclout/cips/discussions/35)
> > 
> > I agree with @tijn
> 
> @harpreet9629 Could you guys take care to name the right Tijn, please? The one with an extra o: @tijno, is the one you need. I understand that typos and small mistakes exists; it's fine, no bad feelings, but for me all these messages are basically spam in my inbox and I have no means to stop it other than hitting the unsubscribe button on each and every pull-request and issue where I'm mentioned by mistake.

Sorry it's my bad

tijno:
@harpreet9629 edit the original post please as you just notified `@tijn` again by quoting your previous reply :)

AdityaChaudhary0005:
Lol i thought this went real
Screenshot from 2021-07-14 19-08-20

harpreet9629:
We can use this PR, because yet there is no approval of auto complete mentions. In being time i am agree with this PR.

AdityaChaudhary0005:
This PR has nothing to do with auto-complete. it will just won't let the mentions to go in lower case automatically.

harpreet9629:
> This PR has nothing to do with auto-complete. it will just won't let the mentions to go in lower case automatically.

Yes, i saw all changes. 😊

maebeam (approved):
This seems like a positive change to me. @lazynina thoughts?

lazynina:
> This seems like a positive change to me. @lazynina thoughts?

Think this is a good change as well!

lazynina (approved):

PR #313, opened 2021-07-19
Cleanup theme logic

maebeam:


maebeam:
Removed the directive from typescript and moved the reference into index.html. This also fixes theme switching which is currently broken.

PR #325, opened 2021-07-27
Rph/nifties

lazynina:


src/app/feed/feed-post/feed-post.component.ts
```diff
@@ -120,6 +149,58 @@ export class FeedPostComponent implements OnInit {
       this.post.RecloutCount = 0;
     }
     this.setEmbedURLForPostContent();
+    if (this.postContent.IsNFT && !this.nftEntryResponses?.length) {
```
maebeam, line 152: Thoughts on tradeoffs around just returning this information in/with the PostEntryResponse?

src/app/feed/feed.component.ts
```diff
@@ -16,8 +16,8 @@ import { Title } from "@angular/platform-browser";
 export class FeedComponent implements OnInit, OnDestroy, AfterViewChecked {
   static GLOBAL_TAB = "Global";
   static FOLLOWING_TAB = "Following";
-  static MARKET_TAB = "Market";
-  static TABS = [FeedComponent.GLOBAL_TAB, FeedComponent.FOLLOWING_TAB];
+  static SHOWCASE_TAB = "💎Showcase💎";
```
maebeam, line 19: Does this look weird without spaces between emoji and text?

src/index.html
```diff
@@ -25,6 +25,7 @@
     <link href="https://fonts.googleapis.com/css?family=Roboto%20Mono:400,500,700" rel="stylesheet" />
     <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet" />
     <link href="/vendor/fontello.css" rel="stylesheet" />
+    <link rel="stylesheet" href="https://unpkg.com/ngx-bootstrap/datepicker/bs-datepicker.css">
```
maebeam, line 28: Let's vendor this

Caddyfile
```diff
@@ -70,5 +70,4 @@ header @html Content-Security-Policy "
     https://open.spotify.com
     https://w.soundcloud.com
     pay.testwyre.com
-    pay.sendwyre.com;
-  frame-ancestor 'self';"
```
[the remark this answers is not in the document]
lazynina: This causes errors in the console, which is why I removed it.

src/app/creator-profile-page/creator-profile-nfts/creator-profile-nfts.component.ts
```diff
@@ -0,0 +1,295 @@
...
+    if (!this.activeTab) {
+      this.isLoading = true;
+      let defaultTab = this.profileBelongsToLoggedInUser()
+        ? CreatorProfileNftsComponent.MY_BIDS
+        : CreatorProfileNftsComponent.MY_GALLERY;
+      this.onActiveTabChange(defaultTab);
+    }
+  }
+
+  getNFTBids(): Subscription {
```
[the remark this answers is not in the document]
lazynina: It makes the page render faster

Caddyfile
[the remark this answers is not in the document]
maebeam: Oh it's supposed to be `frame-ancestors` with an `s` at the end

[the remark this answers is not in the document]
lazynina: :facepalm: good catch. i'll add it back with the `s`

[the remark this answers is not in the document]
lazynina: good catch - this is fixed now.

src/index.html
lazynina: vendored! :+1:

lazynina: Without spaces:
image

With spaces:
image

spaces feels a bit too spaced out IMO

src/app/feed/feed.component.ts
lazynina: Without spaces:
image

With spaces:
image

spaces feels a bit too spaced out IMO

maebeam: Interesting, I definitely prefer the spaces

maebeam: I might also prefer just one diamond at the beginning or end

src/app/feed/feed-post/feed-post.component.ts
lazynina: So the reason I kept it separate is to keep the response size lower when fetching multiple posts. If someone has a few posts with 1000 copies, it can get to be a bit large. Additionally, others building off the API may not need the NFT details to meet their needs.

However, I can't deny that it would simply things if we included it with every PostEntryResponse.

diamondhands0 (approved):

PR #327, opened 2021-07-27
redirect user to inbox thread to send dm from creator profile

lazynina:
image

Clicking "Message User" opens the Inbox component and selects the thread based on the username. 

Also resolves an issue when clicking on the change account selector.


harpreet9629:
Oh that's Great. I did same, but forget to submit pull request, i used button with Buy Button. Anyway its same goal :)

diamondhands0 (approved):

PR #341, opened 2021-07-29
always show NFT details of quoted content if quoted post is NFT, fix styling of quoted content on post thread page

lazynina:
currently on bitclout.com:
image

after updates:
image

- also fixes border when hovering over quoted content

### DIFF
--- a/src/app/feed/feed-post/feed-post.component.html
+++ b/src/app/feed/feed-post/feed-post.component.html
@@ -14,7 +14,7 @@
      - market-feed component -->
 <div
   class="d-flex flex-column"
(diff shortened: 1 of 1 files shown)


maebeam:
Shouldn't the place bid button be right aligned?

lazynina:
For quoted content, especially when there is unlockable content and viewed from a mobile device, there isn't room for it all on one line. 

Regardless, this PR doesn't change the alignment of these elements. I think we can address re-aligning the content of the NFT details on the feed-post component for quoted content in a separate PR.

PR #359, opened 2021-08-04
Ln/jumio

lazynina:


src/app/global-vars.service.ts
```diff
@@ -717,28 +730,63 @@ export class GlobalVarsService {
...
+    this.backendApi
+      .JumioBegin(
+        environment.jumioEndpointHostname,
+        this.loggedInUser?.PublicKeyBase58Check,
+        jumioSuccessRoute,
+        jumioErrorRoute
+      )
+      .subscribe(
+        (res) => {
+          window.open(res.URL);
```
maebeam, line 745: I'd recommend copying `identityService.launch` and specifying the size and location of the new window you're opening.

src/app/global-vars.service.ts
```diff
@@ -717,28 +730,63 @@ export class GlobalVarsService {
...
-    this.logEvent("account : create : launch");
+  launchIdentityFlow(event: string): void {
+    this.logEvent(`account : ${event} : launch`);
     this.identityService.launch("/log-in").subscribe((res) => {
-      this.logEvent("account : create : success");
+      this.logEvent(`account : ${event} : success`);
       this.backendApi.setIdentityServiceUsers(res.users, res.publicKeyAdded);
-      this.updateEverything().subscribe(() => {
-        this.flowRedirect(res.signedUp);
+      let updateFlowFinishedObservable = res.jumioSuccess
```
maebeam, line 767: A comment here about why the order of calls is important would be great

src/app/global-vars.service.ts
```diff
@@ -278,6 +283,11 @@ export class GlobalVarsService {
 
     this.loggedInUser = user;
 
+    // If Jumio callback hasn't returned yet, we need to poll to update the user metadata.
+    if (user.JumioFinishedTime > 0 && !user.JumioReturned) {
+      this.pollLoggedInUserForJumio(user.PublicKeyBase58Check);
```
maebeam, line 288: You need to clear the interval before starting a new one. Otherwise switching accounts back and forth will create tons of intervals

src/app/sign-up/sign-up-get-starter-bitclout/sign-up-get-starter-bitclout.component.html
```diff
@@ -15,6 +15,27 @@
     <div class="fs-18px mt-15px mb-15px">BitClout is a cryptocurrency that's required to use the site.</div>
   </div>
 
+  <div *ngIf="!jumioError && showJumio">
```
maebeam, line 18: I'm confused, if they come back from Identity won't they see this?

src/app/sign-up/sign-up-get-starter-bitclout/sign-up-get-starter-bitclout.component.ts
```diff
@@ -41,11 +44,22 @@ export class SignUpGetStarterBitcloutComponent implements OnInit {
...
-  constructor(public globalVars: GlobalVarsService, private backendApi: BackendApiService) {}
+  constructor(
+    public globalVars: GlobalVarsService,
+    private backendApi: BackendApiService,
+    private route: ActivatedRoute
+  ) {}
 
   ngOnInit(): void {
     this._setScreenToShow();
+    // this.jumioTestCallback();
```
maebeam, line 57: RM

src/app/global-vars.service.ts
```diff
@@ -717,28 +730,63 @@ export class GlobalVarsService {
...
+  launchIdentityFlow(event: string): void {
+    this.logEvent(`account : ${event} : launch`);
     this.identityService.launch("/log-in").subscribe((res) => {
-      this.logEvent("account : create : success");
+      this.logEvent(`account : ${event} : success`);
       this.backendApi.setIdentityServiceUsers(res.users, res.publicKeyAdded);
-      this.updateEverything().subscribe(() => {
-        this.flowRedirect(res.signedUp);
+      let updateFlowFinishedObservable = res.jumioSuccess
+        ? this.backendApi.JumioFlowFinished(
```
maebeam, line 768: Shouldn't this call happen from Identity?

src/app/sign-up/sign-up-get-starter-bitclout/sign-up-get-starter-bitclout.component.html
lazynina: ahh I just need to kill this - I originally had it as part of the app flow but then diamondhands wanted it in identity so it was available to 3rd party nodes

src/app/global-vars.service.ts
lazynina: this complexity will go away once we move all jumio begin / finish logic to identity.

src/app/global-vars.service.ts
lazynina: moving now.

src/app/global-vars.service.ts
lazynina: this interval now gets cleared in the pollLoggedInUserForJumio function

maebeam (approved):

tijno:
see https://github.com/bitclout/backend/issues/101 for bug requiring users that already own their own creator coin to buy it again.

PR #365, opened 2021-08-04
feat: create ability for users to follow creators they are buying

superzordon:
Screen Shot 2021-08-04 at 2 57 05 PM

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -140,7 +140,9 @@ export class AppComponent implements OnInit {
           this.globalVars.youHodlMap[entry.CreatorPublicKeyBase58Check] = entry;
         }
 
-        this.globalVars.defaultFeeRateNanosPerKB = res.DefaultFeeRateNanosPerKB;
+        if (res.DefaultFeeRateNanosPerKB > 0) {
+          this.globalVars.defaultFeeRateNanosPerKB = res.DefaultFeeRateNanosPerKB;
+        }
         this.globalVars.globoMods = res.GloboMods;
         this.ref.detectChanges();
         this.globalVars.loadingInitialAppState = false;
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,6 +1,6 @@
 import { ChangeDetectorRef, Component, HostListener, OnInit } from "@angular/core";
 import { HttpClient } from "@angular/common/http";
-import { BackendApiService } from "./backend-api.service";
+import { BackendApiService, User } from "./backend-api.service";
 import { GlobalVarsService } from "./global-vars.service";
 import { ActivatedRoute, Router } from "@angular/router";
 import { IdentityService } from "./identity.service";
@@ -122,7 +122,7 @@ export class AppComponent implements OnInit {
         }
 
         // Find the loggedInUser in our results
-        const loggedInUser = _.find(res.UserList, { PublicKeyBase58Check: loggedInUserPublicKey });
+        const loggedInUser: User = _.find(res.UserList, { PublicKeyBase58Check: loggedInUserPublicKey });
 
         // Only update if things have changed to avoid unnecessary DOM manipulation
         if (!_.isEqual(this.globalVars.loggedInUser, loggedInUser)) {
@@ -170,6 +170,8 @@ export class AppComponent implements OnInit {
         this.globalVars.diamondLevelMap = res.DiamondLevelMap;
         this.globalVars.showProcessingSpinners = res.ShowProcessingSpinners;
         this.globalVars.showBuyWithUSD = res.HasWyreIntegration;
+        this.globalVars.showJumio = res.HasJumioIntegration;
+        this.globalVars.jumioBitCloutNanos = res.JumioBitCloutNanos;
         // Setup amplitude on first run
         if (!this.globalVars.amplitude && res.AmplitudeKey) {
           this.globalVars.amplitude = require("amplitude-js");
--- a/src/app/follow-button/follow-button.component.ts
+++ b/src/app/follow-button/follow-button.component.ts
@@ -12,6 +12,7 @@ import { Input } from "@angular/core";
 import { FollowChangeObservableResult } from "../../lib/observable-results/follow-change-observable-result";
 import { Subscription } from "rxjs";
 import { CanPublicKeyFollowTargetPublicKeyHelper } from "../../lib/helpers/follows/can_public_key_follow_target_public_key_helper";
+import { FollowService } from "./follow.service";
 
 @Component({
   selector: "follow-button",
@@ -31,18 +32,19 @@ export class FollowButtonComponent implements OnInit, OnDestroy {
   // Is the logged in user currently following the target person?
   isFollowing: boolean;
   appData: GlobalVarsService;
+  followService: FollowService;
   createFollowTxnBeingCalled = false;
   followChangeSubscription: Subscription;
   changeRef: ChangeDetectorRef;
 
   unfollow(event) {
     event.stopPropagation();
-    this._toggleFollow(false);
+    this.followService._toggleFollow(false);
   }
 
   follow(event) {
     event.stopPropagation();
-    this._toggleFollow(true);
+    this.followService._toggleFollow(true);
   }
 
   canLoggedInUserFollowTargetPublicKey() {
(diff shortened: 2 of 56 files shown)


src/app/follow-button/follow-button.component.ts
```diff
@@ -222,6 +105,13 @@ export class FollowButtonComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit() {
-    this.isFollowing = this._isLoggedInUserFollowing();
+    this.followService = new FollowService(
```
maebeam, line 108: No need to construct this service. Just add it as an argument in the constructor and angular will handle it. Instead, the methods in followService need to take `this.followedPubKeyBase58Check` as a param

src/styles.scss
```diff
@@ -2467,3 +2467,8 @@ input {
     }
   }
 }
+
+.checkbox-warning-filled [type="checkbox"][class*='filled-in']:checked+label:after {
+  border-color: #FF8800 !important;
+  background-color: #FF8800 !important;
```
lazynina, line 2473: these should be set based on the theme. use var(--background) I think

src/app/follow-button/follow.service.ts
```diff
@@ -0,0 +1,141 @@
+import { ChangeDetectorRef, Injectable} from "@angular/core";
```
lazynina, line 1: we should probably move this to live near the rest of the services instead of burying it in the follow-button directory

src/app/trade-creator-page/trade-creator-table/trade-creator-table.component.scss
```diff
@@ -0,0 +1,51 @@
+*, *:after, *:before {box-sizing: border-box;}
```
lazynina, line 1: we have stayed away from putting styling in these scss files located in directories. There could be a case made that it is nice to keep it separate, but generally diamondhands and redpartyhat have wanted all styling to be in the route styles.scss

lazynina (changes requested):
some code style changes requested. I need to test this out locally a bit more.

lazynina:
We should probably hide the checkbox if the user is buying their own coin.

src/lib/services/follow/follow.service.ts
```diff
@@ -0,0 +1,131 @@
...
+        followedPubKeyBase58Check,
+        !isFollow /*isUnfollow*/,
+        this.appData.feeRateBitCloutPerKB * 1e9
+      )
+      .subscribe(
+        (response) => {
+          this._handleSuccessfulFollowTxn(isFollow, followedPubKeyBase58Check);
+          this._notifyFollowChangeObservers(followedPubKeyBase58Check);
+        },
+        (response) => {
```
lazynina, line 52: let's call this error since we call the response above `response`

src/app/trade-creator-page/trade-creator-preview/trade-creator-preview.component.ts
```diff
@@ -123,6 +125,13 @@ export class TradeCreatorPreviewComponent implements OnInit {
 
           // This will update the user's balance.
           this.appData.updateEverything(response.TxnHashHex, this._creatorCoinSuccess, this._creatorCoinFailure, this);
+
+          if (
+            this.creatorCoinTrade.followCreator &&
+            !this.followService._isLoggedInUserFollowing(this.creatorCoinTrade.creatorProfile.PublicKeyBase58Check)
+          ) {
+            this.followService._toggleFollow(true, this.creatorCoinTrade.creatorProfile.PublicKeyBase58Check);
+          }
```
lazynina, line 134: Okay so i think we need to call updateEverything after the follow transaction is completed as well or at least update the logged in user to show that the user is following the public key. If I buy a creator coin and have the Follow checkbox checked, I get my creator coins and the follow transaction goes through, but if I go to buy more of the creator coin, I still see the checkbox.

lazynina (changes requested):

src/app/trade-creator-page/trade-creator-preview/trade-creator-preview.component.ts
maebeam: I thought `_toggleFollow` updated the local cache of who you are following

lazynina: you're right - it does do that. I think something strange is happening since we're calling `updateEverything` around the same time as toggle follow. Sometimes after I buy a creator with the checkbox checked, I'll still have the option to follow. However, if I perform a refresh, it doesn't show up. My guess is that we've got some race condition where we get undesirable behavior when updateEverything or the follow transaction gets processed in a slightly different order.

lazynina: my guess is that we should call updateEverything once after toggling the follow instead of at the same time. In the event that the checkbox is not checked or the user already follows the creator, we can just call updateEverything

src/app/trade-creator-page/trade-creator-preview/trade-creator-preview.component.ts
```diff
@@ -121,8 +123,30 @@ export class TradeCreatorPreviewComponent implements OnInit {
...
+              });
+          } else {
+            // This will update the user's balance.
+            this.appData.updateEverything(
+              response.TxnHashHex,
+              this._creatorCoinSuccess,
+              this._creatorCoinFailure,
+              this
+            );
+          }
```
lazynina, line 149: we may be able to clean this up a bit (note: `import { of } from "rxjs";`  at the top of this file):
```
const observable = this.creatorCoinTrade.followCreator &&
            !this.followService._isLoggedInUserFollowing(this.creatorCoinTrade.creatorProfile.PublicKeyBase58Check) ? this.followService._toggleFollow(true, this.creatorCoinTrade.creatorProfile.PublicKeyBase58Check) : of(null);
observable.add(() => {
this.appData.updateEverything(
              response.TxnHashHex,
              this._creatorCoinSuccess,
              this._creatorCoinFailure,
              this
});
```

src/app/trade-creator-page/trade-creator-table/trade-creator-table.component.html
```diff
@@ -96,25 +96,36 @@
...
-          class="fas fa-info-circle text-greyC fs-15px global__tooltip-icon"
+          class="fas fa-info-circle text-grey5 fs-15px global__tooltip-icon"
           matTooltipClass="global__mat-tooltip global__mat-tooltip-font-size"
           [matTooltip]="TradeCreatorTableComponent.FOUNDER_REWARD_EXPLANATION"
           #tooltip="matTooltip"
         ></i>
       </span>
     </div>
   </div>
+  <div *ngIf="!userFollowingCreator" class="py-1 border-bottom text-muted d-flex text-grey9 follow-on-buy">
```
lazynina, line 120: add to the *ngIf something to check that  the loggedInUser isn't purchasing themselves.

src/app/trade-creator-page/trade-creator-table/trade-creator-table.component.ts
```diff
@@ -13,6 +13,7 @@ export class TradeCreatorTableComponent {
 
   @Input() displayForCreatorForm: boolean = false;
   @Input() creatorCoinTrade: CreatorCoinTrade;
+  @Input() userFollowingCreator: boolean;
```
lazynina, line 16: instead of making this an input, should we just add the follow service in the constructor and compute this on init?

lazynina (changes requested):

lazynina:
also, this element shouldn't appear when I try to sell a creator's coins:
image

src/app/app.component.ts
```diff
@@ -143,7 +143,6 @@ export class AppComponent implements OnInit {
         if (res.DefaultFeeRateNanosPerKB > 0) {
           this.globalVars.defaultFeeRateNanosPerKB = res.DefaultFeeRateNanosPerKB;
         }
-        this.globalVars.defaultFeeRateNanosPerKB = res.DefaultFeeRateNanosPerKB;
```
maebeam, line 146: Intended?

src/app/trade-creator-page/trade-creator/trade-creator.component.ts
```diff
@@ -76,6 +77,8 @@ export class TradeCreatorComponent implements OnInit {
     this.creatorCoinTrade.setTradeType(tab);
     this.creatorCoinTrade.selectedCurrency = this.creatorCoinTrade.defaultCurrency();
 
+    console.log("Here is the trade type of the parent");
+    console.log(this.creatorCoinTrade.tradeType);
```
lazynina, line 81:
```suggestion
```

src/app/trade-creator-page/trade-creator-table/trade-creator-table.component.ts
```diff
@@ -14,7 +15,12 @@ export class TradeCreatorTableComponent {
   @Input() displayForCreatorForm: boolean = false;
   @Input() creatorCoinTrade: CreatorCoinTrade;
 
+  hideFollowPrompt = true;
   TradeCreatorTableComponent = TradeCreatorTableComponent;
+  appData: GlobalVarsService;
+  buyVerb = CreatorCoinTrade.BUY_VERB
```
lazynina, line 21:
```suggestion
  buyVerb = CreatorCoinTrade.BUY_VERB;
```

src/app/trade-creator-page/trade-creator-table/trade-creator-table.component.html
```diff
@@ -96,25 +96,45 @@
...
         ></i>
       </span>
     </div>
   </div>
+  <div
+    *ngIf="
+      !(
+        this.followService._isLoggedInUserFollowing(this.creatorCoinTrade.creatorProfile.PublicKeyBase58Check) ||
+        this.appData.loggedInUser.PublicKeyBase58Check === this.creatorCoinTrade.creatorProfile.PublicKeyBase58Check ||
+        this.creatorCoinTrade.tradeType !== buyVerb
```
lazynina, line 125:
```suggestion
        followService._isLoggedInUserFollowing(creatorCoinTrade.creatorProfile.PublicKeyBase58Check) ||
        appData.loggedInUser.PublicKeyBase58Check === creatorCoinTrade.creatorProfile.PublicKeyBase58Check ||
        creatorCoinTrade.tradeType !== buyVerb
```

lazynina (approved):

PR #366, opened 2021-08-04
fix: removed page refresh when nft bid is made

superzordon:
Removed page refreshes when bid placed, NFT sold, NFT sold with unlockable content, and auction closed

### DIFF
--- a/src/app/nft-post-page/nft-post/nft-post.component.ts
+++ b/src/app/nft-post-page/nft-post/nft-post.component.ts
@@ -218,6 +218,11 @@ export class NftPostComponent {
     this.globalVars.loggedInUser.BlockedPubKeys[blockedPubKey] = {};
   }
 
+  afterNftBidPlaced() {
+    this.loading = true;
+    this.refreshPosts();
+  }
+
   sellNFT(): void {
     if (this.sellNFTDisabled) {
       return;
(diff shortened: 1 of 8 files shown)


maebeam:
We need to pull fresh data so you can see your bid you've placed

lazynina:
@superzordon we need to handle the case where a user submits a bid from the nft-post-page. It looks like we only update the feed-post's NFT entries when that happens. I'd suggest that we emit an event from feed-post in the subscription to the onHide event for the `PlaceBidModalComponent`. The NFT Post Page component should bind to that event and then perform the following code:
```
this.loading = true;
this.refreshPosts();
```

src/app/nft-post-page/nft-post/nft-post.component.ts
```diff
@@ -218,6 +218,11 @@ export class NftPostComponent {
     this.globalVars.loggedInUser.BlockedPubKeys[blockedPubKey] = {};
   }
 
+  afterNftBidPlaced() {
+    this.loading = true;
+    this.refreshPosts();
```
lazynina, line 223: we only need to call `this.refreshBidData()` actually - sorry!

lazynina (changes requested):
after line 585, add `this.nftBidPlaced.emit()` to trigger the nft post page to refresh its bids.

src/app/nft-post-page/nft-post/nft-post.component.ts
```diff
@@ -1,4 +1,4 @@
-import { ChangeDetectorRef, Component } from "@angular/core";
+import {ChangeDetectorRef, Component, ViewChild} from "@angular/core";
```
lazynina, line 1:
```suggestion
import { ChangeDetectorRef, Component, ViewChild } from "@angular/core";
```

lazynina (approved):
nice work!

PR #375, opened 2021-08-10
fix: lowest bid shows highest bid

tijno:
Lowest Bid bug fixed.

20210810-eL4BhOWp

Eg:

https://bitclout.com/nft/ac20e9cb5a7e6cec8955880f9f563ea6853cf5a963fb34aa6eb5f84e38195331?nftTab=for_sale&tab=posts&feedTab=Following

Fix live on tijn.club

Eg:

https://tijn.club/nft/ac20e9cb5a7e6cec8955880f9f563ea6853cf5a963fb34aa6eb5f84e38195331?nftTab=for_sale&tab=posts&feedTab=Following

### DIFF
--- a/src/app/feed/feed-post/feed-post.component.html
+++ b/src/app/feed/feed-post/feed-post.component.html
@@ -478,11 +478,14 @@
       <div>
     </div>
     <div *ngIf="showExpandedNFTDetails && (highBid || lowBid)" [ngClass]="{'pt-15px': globalVars.isMobile()}">
-      <div *ngIf="highBid != null">
+      <div *ngIf="highBid != 0">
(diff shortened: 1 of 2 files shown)


lazynina:
@tijno this "From" attribute is not meant to show the lowest bid across all serial numbers, rather it is meant to show the lowest amount at which you could place a bid and be in the lead for any serial number. 

For a 1-of-1 this field isn't super helpful, so maybe we want to hide the "From" field in these cases.

tijno:
ah got it! of course. thx @lazynina 

i can amend pr to hide from (lowBid) in case lowBid=highBid

tijno:
Found some edge cases where this does not work but cant put this back in draft so will close it and re-open when i got those sorted.

PR #376, opened 2021-08-10
fix: Fixed number formatting for displayed USD amounts

superzordon:
Before: 
image
image
image

After:
image
image
image


lazynina:
fixed by #374

PR #377, opened 2021-08-10
don't fetch the profile picture if the logged in user doesn't have a profile yet

lazynina:


src/app/avatar/avatar.directive.ts
```diff
@@ -12,6 +12,7 @@ export class AvatarDirective implements OnChanges {
   constructor(private globalVars: GlobalVarsService, private backendApi: BackendApiService, private el: ElementRef) {}
 
   setAvatar() {
+    console.log(this.avatar);
```
lazynina, line 15:
```suggestion
```

maebeam (approved):

PR #380, opened 2021-08-11
CLOUT Diamonds

diamondhands0:
Add support for CLOUT diamonds PLUS...

* Allow anon diamonds to be given. This means we don't require a profile
to be created prior to giving a diamond.

* Handle anon diamonds gracefully on the Diamonds tab by grouping the
anon amounts together into one row.

* Don't show founder reward until user has bought some of their coin.

* Don't ask for confirmation on $5 diamonds because it's annoying.

* Update copy in various places to support CLOUT diamonds and anon
diamonds.


maebeam (approved):

lazynina (approved):
:+1:

Issue #394, opened 2021-08-16
False Diamond triggers


iPaulPro:
There's an issue with the hover detection in the new Diamond slider.

<img src="https://user-images.githubusercontent.com/354227/129568188-57272d73-9de9-4341-8647-460b1bc670c4.gif" width="360">

Lots of reports of accidental diamond giving, and with the removal of the confirmation for $5 level this is bound to happen frequently.

attn @superzordon

seanslater:
I just mocked up this concept.... instead of horizontal,  it could go vertical.

You'd move up the list as you wanted to give more diamonds.  This would make it a more deliberate action and have the default as 1 diamond.

![vertical](https://user-images.githubusercontent.com/378385/129574323-e841a0d3-771f-499c-a660-72b7c34147d8.jpg)

tijno:
Added a gif to show the issue is also present on the other values but then is gone again in a small zone around the border of the box. 

![20210816-kv3js4Uw](https://user-images.githubusercontent.com/69529928/129574639-320865ec-c6c7-402d-ba95-61e77fc930df.gif)


I like @seanslater idea of vertical slider.

tijno:
This may be related to - when I click the diamond button it implies im giving level 3, until the spinner finishes.

![20210816-8G3lEjWq](https://user-images.githubusercontent.com/69529928/129576690-99d13e67-d217-4e91-9570-fb7c7032c487.gif)

superzordon:
Thanks for flagging. Going to reduce the size of the hoverable select box and add a delay to the pop-up. Here's what it will look like: 

https://user-images.githubusercontent.com/88362450/129577003-795e4698-b24a-41f4-928a-e6325145d24c.mov

tijno:
@superzordon 

1) can you show what happens when just clicking the diamond button without moving mouse?

2) can you try and show the behaviour paul and i demonstrated is resolved by slowing down the mouse movement? Diamond selector should not select any diamonds until mouse cursor is within the popup select box

tijno:
for info - https://github.com/bitclout/frontend/pull/395

tijno:
Looking good. Ill check now.

tijno:
@superzordon Safari on macOs still seems to have the issue

![20210816-ESq2cJAV](https://user-images.githubusercontent.com/69529928/129592327-3f2baa53-5faa-4716-9ff8-e72f01a0ebe4.gif)

[...]

PR #395, opened 2021-08-16
Add hover delay to diamond pop-up, fix issue with logged out users

superzordon:
https://user-images.githubusercontent.com/88362450/129578727-df5955d5-c5e7-4b28-822b-eff3f460c598.mov


maebeam (approved):

tijno:
@superzordon doesn't look like the issue highlighted in #394 is solved.

20210816-4eTaM4RB

tijno:
Im on safari mac os - not sure that make a difference but that last commit didnt seem to have made a difference.

20210816-MtEbrOrt

Also - with the increased delay is it worth showing the hover indicator on the diamond like we have on likes/reclout ?

PR #398, opened 2021-08-18
Update Content Security Policy for IPFS

mubashariqbal:
For an upcoming NFT project, I wish to host my image files using IPFS.  

This update enables images hosted on IPFS and via Pinata (https://www.pinata.cloud/) to be displayed on BitClout.com currently blocked by the Content Security Policy.

ipfs.io: default ipfs url
gateway.ipfs.io: default ipfs gateway url
gateway.pinata.cloud: default pinata ipfs gateway
*.mypinata.cloud: custom gateway urls enabled by pinata

List of IPFS gateways:
https://ipfs.github.io/public-gateway-checker/

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -51,7 +51,7 @@ header @html Content-Security-Policy "
     'unsafe-inline'
     https://fonts.googleapis.com
     https://cdn.jsdelivr.net/npm/bootstrap@4.6.0/dist/css/bootstrap.min.css;
-  img-src 'self' data: i.imgur.com images.bitclout.com quickchart.io arweave.net *.arweave.net;
+  img-src 'self' data: i.imgur.com images.bitclout.com quickchart.io arweave.net *.arweave.net ipfs.io gateway.ipfs.io gateway.pinata.cloud *.mypinata.cloud;
   font-src 'self'
     https://fonts.googleapis.com
     https://fonts.gstatic.com


seanslater:
Would it be worth adding the Cloudflare IPFS gateway too? I've found it to be much faster.

developers.cloudflare.com/distributed-web/ipfs-gateway

mubashariqbal:
> Would it be worth adding the Cloudflare IPFS gateway too? I've found it to be much faster.
> 
> developers.cloudflare.com/distributed-web/ipfs-gateway

I'm not using Cloudflare in this instance so didn't include in my PR, but happy to update if others thing it would be useful.

redpartyhat (approved):

redpartyhat:
> > Would it be worth adding the Cloudflare IPFS gateway too? I've found it to be much faster.
> > developers.cloudflare.com/distributed-web/ipfs-gateway
> 
> I'm not using Cloudflare in this instance so didn't include in my PR, but happy to update if others thing it would be useful.

I see no harm in adding the CF gateway for more flexibility.

lazynina (approved):
:rocket:

mubashariqbal:
Thanks for the quick merge!  Is there anyway to be notified when this deployed?

tijno:
I have an alert @mubashariqbal the new frontend goes live - can ping you when I see it, or add you to telegram group chat where the alert comes in

mubashariqbal:
> I have an alert @mubashariqbal the new frontend goes live - can ping you when I see it, or add you to telegram group chat where the alert comes in

Would love to be added to the telegram group!

maebeam:
Can we replace pinata with cloudflare? I trust cloudflare more than pinatas

mubashariqbal:
> Can we replace pinata with cloudflare without issue? I trust cloudflare more than pinatas

The reason I chose to use Pinata, is because you don't need to setup your own server to host/pin the images, but I'd be happy with either if you have a strong preference.

redpartyhat:
I guess one downside of pinata is that you are dependent on their service not going down in order to serve the images, correct?  Slash, on the NFT minter paying for the service? https://docs.pinata.cloud/gateways/dedicated-gateways

mubashariqbal:
> I guess one downside of pinata is that you are dependent on their service not going down in order to serve the images, correct? Slash, on the NFT minter paying for the service? https://docs.pinata.cloud/gateways/dedicated-gateways

Once it's on IPFS you can serve from any gateway I suppose, but you're always reliant on something to keep it pinned. Either your own server or something line Pinata.

maebeam:
> The reason I chose to use Pinata, is because you don't need to setup your own server to host/pin the images, but I'd be happy with either if you have a strong preference.

I don't follow. From my understanding of IPFS, you don't need to setup your own server to have CloudFlare serve your file. As long as there is an IPFS node on the network with your file CloudFlare will find it, cache it, and serve it. For example, upload on Pinata to get a CID and then serve with CF. Please let me know if I'm misunderstanding something here.

> I guess one downside of pinata is that you are dependent on their service not going down in order to serve the images, correct? Slash, on the NFT minter paying for the service? https://docs.pinata.cloud/gateways/dedicated-gateways

Not quite. Ideally, the chain would store the IPFS CID and not a URL. Then, any node could use any gateway to try and serve the file. If no nodes have the file pinned then you're screwed. If anything, bitclout.com should have a dedicated gateway and we should pin IPFS files that users post from our frontend.

mubashariqbal:
> > The reason I chose to use Pinata, is because you don't need to setup your own server to host/pin the images, but I'd be happy with either if you have a strong preference.
> 
> I don't follow. From my understanding of IPFS, you don't need to setup your own server to have CloudFlare serve your file. As long as there is an IPFS node on the network with your file CloudFlare will find it, cache it, and serve it. For example, upload on Pinata to get a CID and then serve with CF. Please let me know if I'm misunderstanding something here.

Yes I was referring to the pinning, as long as it still on IPFS any gateway can serve it.

> > I guess one downside of pinata is that you are dependent on their service not going down in order to serve the images, correct? Slash, on the NFT minter paying for the service? https://docs.pinata.cloud/gateways/dedicated-gateways
> 
> Not quite. Ideally, the chain would store the IPFS CID and not a URL. Then, any node could use any gateway to try and serve the file. If no nodes have the file pinned then you're screwed. If anything, bitclout.com should have a dedicated gateway and we should pin IPFS files that users post from our frontend.

That would be the ideal situation, but I'm guessing that is a long term CIP required change. 

Passing the URL and storing it in the ImageURLs for a post was a quick fix for my immediate need.

redpartyhat:
@mubashariqbal, it seems like the best option here is to add the CloudFlare gateway only.  This will force everyone to use the fastest and most well-known gateway provider, which will be good for performance and make it easier to clean up if we want to parse the links and send them through a dedicated bitclout.com gateway in the future.  I think with this set up you can still use Pinata or any other provider as a pinning service but provide the CloudFlare link in your post instead.

Does this make sense to you?  If so, would you mind pushing a fix that removes pinata and replaces it with CloudFlare?

mubashariqbal:
> @mubashariqbal, it seems like the best option here is to add the CloudFlare gateway only. This will force everyone to use the fastest and most well-known gateway provider, which will be good for performance and make it easier to clean up if we want to parse the links and send them through a dedicated bitclout.com gateway in the future. I think with this set up you can still use Pinata or any other provider as a pinning service but provide the CloudFlare link in your post instead.
> 
> Does this make sense to you? If so, would you mind pushing a fix that removes pinata and replaces it with CloudFlare?

Yes this makes perfect sense.

Since this PR was already merged made a new one for the updates.  #401

PR #402, opened 2021-08-18
set exchange rate from app state

lazynina:


src/app/app.component.ts
```diff
@@ -166,6 +166,12 @@ export class AppComponent implements OnInit {
     this.backendApi
       .GetAppState(this.globalVars.localNode, this.globalVars.loggedInUser?.PublicKeyBase58Check)
       .subscribe((res: any) => {
+        this.globalVars.ExchangeUSDCentsPerBitClout = res.USDCentsPerBitCloutExchangeRate;
```
diamondhands0, line 169: Does this take the max(reserve price, blockchain.com price) ?

lazynina: Yes - this value is produced by fes.GetExchangeBitCloutPrice(), the same function used by GetExchangeRate to set the value of USDCentsPerBitCloutExchangeRate

PR #404, opened 2021-08-19
WIP: Ln/testnet tutorial

lazynina:


src/app/app.component.ts
```diff
@@ -195,6 +206,7 @@ export class AppComponent implements OnInit {
   }
 
   _updateEverything = (
+    refreshAllUsers: boolean = false,
```
maebeam, line 209: I'd prefer not to add more complexity to this crazy method. Could you share why this is needed?

maebeam (approved):

src/app/app.component.ts
lazynina: it's really slow to call updateEverything on the users you have logged in. If we only care about refreshing the loggedInUser (which 99% of the time is all we care about), then this GetUsersStateless call is much faster. We end up calling updateEverything after each step in the tutorial basically

lazynina: Perhaps a better solution here is that we only ever refresh the logged in user's data and only hit GetUsersStateless for another public key when we switch accounts. On app load, we find a way to show all users in the change account selector? maybe a lighter weight function that just returns the username for each public key if it exists.

maebeam: I like the idea of a different call to `GetUsersStateless` that populates the account selector. We don't even need to optimize to just return usernames.

lazynina: I think I can just set SkipForLeaderboard to true to avoid all the expensive stuff.

PR #406, opened 2021-08-19
fix: move hover bridge from before attribute on button to after attri…

superzordon:
…bute on diamond box. Expand width of hover box so as to not lose focus when user moves mouse too far left or right

Before:
ezgif com-gif-maker (1)

After:
ezgif com-gif-maker

(Green background added for gif demonstration)


maebeam (approved):
Nice!

PR #407, opened 2021-08-20
Remove deprecated stake entry logic

lazynina:
remove stake-entry related code from feed-post component and from PostEntryResponse class


maebeam:
Thanks!

PR #415, opened 2021-08-24
add skip button to each step of tutorial for admins and users who have elected to go into tutorial

lazynina:


AdityaChaudhary0005:
This was life saver Nina 👏

PR #427, opened 2021-08-27
Add $CLOUT tooltips to CC holdings/price on wallet page

netpro2k:
With recent price volatility it has been hard to tell what is happening to the value of CC holdings. It would be nice to be able to see values in $CLOUT instead of USD. 

This PR adds a simple system tooltip (title attribute) to the wallet total, CC price, and CC cashout values to show the equivalent values in $CLOUT.

clout-tooltip

Ultimately this is something that should probably happen site wide anywhere a price is displayed (along with the ability to change the default denomination to $CLOUT or BTC) but that is a much bigger change and probably involves some design decisions.

This is the lowest effort, highest impact, version of that.

### DIFF
--- a/src/app/wallet/wallet.component.html
+++ b/src/app/wallet/wallet.component.html
@@ -6,7 +6,10 @@
   <top-bar-mobile-navigation-control class="mr-15px d-lg-none d-inline-block"></top-bar-mobile-navigation-control>
   <div class="d-flex align-items-center w-100 justify-content-between">
     <div>Wallet</div>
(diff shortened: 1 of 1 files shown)


AdityaChaudhary0005:
great work @netpro2k ! do you think if we can display it instead of hover event ?

netpro2k:
> great work @netpro2k ! do you think if we can display it instead of hover event ?
Yep I do think there are better solutions to this with some further design changes but that would probably require some more back and forth. This feels obviously better than what we have now without having to really sacrifice anything. I think a site wide solution to this should be figured out, but that's a bigger discussion.

PR #430, opened 2021-08-31
pass referral code to identity and get referral amount

lazynina:


src/environments/environment.prod.ts
```diff
@@ -1,7 +1,7 @@
 export const environment = {
   production: true,
   uploadImageHostname: "bitclout.com",
-  jumioEndpointHostname: "bitclout.com",
+  jumioEndpointHostname: "localhost:18001",
```
redpartyhat, line 4: Is this correct? It seems like "environment.prod.ts" should point to bitclout.com for the jumio endpoint host

lazynina: this is just for testing purposes - we'll flip this back once we're done testing.

PR #433, opened 2021-09-06
Displays last accepted bid price on single-serial NFT feed posts

iPaulPro:
Resale generally indicates demand, but right now it's difficult to distinguish an NFT that is being resold from one that has been passed over due to lack of interest.

This updates the feed post component to show the "last price" on _of 1_ NFTs.

### DIFF
--- a/src/app/feed/feed-post/feed-post.component.ts
+++ b/src/app/feed/feed-post/feed-post.component.ts
@@ -95,6 +95,7 @@ export class FeedPostComponent implements OnInit {
   @Input() nftCollectionHighBid = 0;
   @Input() nftCollectionLowBid = 0;
   @Input() isForSaleOnly: boolean = false;
+  @Input() nftLastAcceptedBidAmountNanos: number;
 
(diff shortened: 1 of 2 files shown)


maebeam (approved):

src/app/feed/feed-post/feed-post.component.ts
```diff
@@ -95,6 +95,7 @@ export class FeedPostComponent implements OnInit {
   @Input() nftCollectionHighBid = 0;
   @Input() nftCollectionLowBid = 0;
   @Input() isForSaleOnly: boolean = false;
+  @Input() nftLastAcceptedBidAmountNanos: number;
```
lazynina, line 98: this doesn't look like an input to the component, but rather a simple attribute since the parent component won't be passing this value. I think we can convert this line to the below
```suggestion
  nftLastAcceptedBidAmountNanos: number;
```

lazynina (changes requested):

lazynina (approved):

PR #436, opened 2021-09-13
CloutCast -- Total Frontend Integration with CloutCast APIs

Zyther:
The following PR modifies the frontend to support CloutCast. This PR is purely additive, does not touch backend code, and most if not all calls are non-blocking to the rest of the frontend's main functionality. Due to these points, a CIP was not created. 

In addition to these points, this PR directly implements the "Sponsored Posts" bullet point in BitClout's white paper. 
image

"Cast Post" has been added to feed-post's action menu, which brings up a new modal to create a Cast.
Casts in the sidebar brings up a "Messages" like page to view casts available to execute. 

This PR is mobile, and desktop ready. 

All screens were made to function and operate like Bitclout's frontend. 

For any questions, concerns, or comments, please reach me directly. (smartalec)


jsondevlin:
A little demo of you’d like to see it live: https://vimeo.com/604146038

tijno:
First off - amazing work @Zyther and @jsondevlin 

As this has had no replies for almost 24 hours - starting things off - some of which I have already shared with you.

Not had a chance to look over the code yet - so apologies if some of this is covered by that.

1 - probably should have a cip before PR as I would say this is a large change.
2 - not all nodes may want this so suggest its disabled by default and nodes can manually "enable" the integration via a switch in the admin panel
3 - i presume using this is at no cost to the node operator
4 - is revenue shared with the node operator - eg like you do with cloutfeed
5 - sposored reclouts/quotes/posts that have been paid for should have some identifier on chain and in backend api which allows front ends to display it was sponsored / paid for.

jsondevlin:
@tijno you're the man! Thank you so much for giving this some thought and taking the time to comment.

1- Sure thing we'll create a CIP so the community can engage with the idea!

2- Feature Flag is a good idea, will let core team add this if they choose as its a bit intrusive.

3- Correct, no cost to node operators.

4- This PR as is does not include revenue share. We're always open to discussing 3rd party integration agreements to find a fair revenue share agreement similar to what we've done with Cloutfeed.

5- We 100% want to find the best way to support this so that any frontend that displays promoted content can add an indicator. This is a problem today, not only when this PR is merged, as content is being promoted today. We're open to thoughts on how to accomplish this best for the ecosystem, but our first thought would be to provide a public endpoint to return this data (future state).

Zyther:
@tijno @jsondevlin 
https://github.com/bitclout/cips/discussions/151

cyrusabrahim:
Thank you Cloutcast team for putting this together. I’m all for it. I agree with proposed changes to PR by @tijno.

Zyther:
Thoughts, @diamondhands0  @maebeam  @lazynina  @bluepartyhat @redpartyhat ?

Zyther:
This is now up to date with DESO changes.

PR #437, opened 2021-09-14
fix: change "skip" to "exit" as skip implies skipping a step.

tijno:
Ive seen this feedback a few times, and run into this myself. 

The "Skip" button in the tutorial implies skipping a step.

So i changed it to "Stop" instead.

20210914-qZVeCnRY


tijno:
Added option to continue with tutorial 

20210914-12jLerTT

ClayOglesby:
"Exit" would make even more sense in my opinion.

Screen Shot 2021-09-15 at 3 27 51 AM

PR #438, opened 2021-09-14
FIX: enter keydown function name in remove verification and 1-of-1 NFT bid selection bug

lazynina:
fix function on enter keydown in remove verification badge, fix setting a bid as selected when selecting a winner for 1-of-1 NFTs


superzordon (approved):
🔥

PR #445, opened 2021-09-15
Buy clout with eth

maebeam:


src/app/backend-api.service.ts
```diff
@@ -69,8 +74,15 @@ export class BackendRoutes {
   static RoutePathGetNextNFTShowcase = "/api/v0/get-next-nft-showcase";
   static RoutePathGetNFTCollectionSummary = "/api/v0/get-nft-collection-summary";
   static RoutePathGetNFTEntriesForPostHash = "/api/v0/get-nft-entries-for-nft-post";
+
+  // Verify
```
redpartyhat, line 78: Nit: since "Verify" is already used, we should either move this endpoint under the other header or change the title of this section

src/app/buy-bitclout-page/buy-bitclout/buy-bitclout.component.ts
```diff
@@ -497,8 +488,16 @@ export class BuyBitcloutComponent implements OnInit {
 
   ngOnInit() {
     window.scroll(0, 0);
-    this.showAreYouReady =
-      this.appData != null && this.appData.loggedInUser != null && this.appData.loggedInUser.BalanceNanos === 0;
+
+    // Add extra tabs
+    if (this.globalVars.showBuyWithUSD) {
+      this.buyTabs.push(BuyBitcloutComponent.BUY_WITH_USD);
+      this.activeTab = BuyBitcloutComponent.BUY_WITH_USD;
```
redpartyhat, line 495: The default state when you have btc, wyre, and eth enabled looks weird now since BUY_WITH_USD is in the middle and selected.  Maybe make BUY_WITH_USD the first tab when it's enabled?
image

PR #447, opened 2021-09-18
Add min bid to single nfts if greater then 0

tijno:
Too often get caught out when I see low last bid, click on make bid, and turns out min bid is way higher.

So added to @iPaulPro recent change to also show min bid if its greater than 0.

20210918-GVkvgdxn

### DIFF
--- a/src/app/feed/feed-post/feed-post.component.ts
+++ b/src/app/feed/feed-post/feed-post.component.ts
@@ -96,6 +96,7 @@ export class FeedPostComponent implements OnInit {
   @Input() nftCollectionLowBid = 0;
   @Input() isForSaleOnly: boolean = false;
   nftLastAcceptedBidAmountNanos: number;
+  nftMinBidAmountNanos: number;
 
   @Input() showNFTDetails = false;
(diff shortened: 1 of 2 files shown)


src/app/feed/feed-post/feed-post.component.ts
```diff
@@ -197,6 +198,9 @@ export class FeedPostComponent implements OnInit {
         this.lowBid = _.minBy(this.availableSerialNumbers, "HighestBidAmountNanos")?.HighestBidAmountNanos || 0;
         if (this.nftEntryResponses.length === 1) {
           this.nftLastAcceptedBidAmountNanos = this.nftEntryResponses[0].LastAcceptedBidAmountNanos;
+          if (this.nftEntryResponses[0].MinBidAmountNanos>0) {
```
lazynina, line 201:
```suggestion
          if (this.nftEntryResponses[0].MinBidAmountNanos > 0) {
```

lazynina (changes requested):

lazynina (approved):

maebeam (approved):

PR #460, opened 2021-09-22
Ln/native video

lazynina:
- Do we want to add a new icon in the feed-create-post component to distinguish between adding an image and a video or find an icon that matches both?

Uploading progress bar:
image

Video Processing (file has been uploaded, but is being processed by cloudflare to optimize streaming)
image

Once video is ready to stream, it will appear in the feed-create-post-component
image

Video in post:
image


src/app/feed/feed-create-post/feed-create-post.component.ts
```diff
@@ -234,8 +244,64 @@ export class FeedCreatePostComponent implements OnInit {
...
+      //   this.globalVars._alertError("File is too large. Please choose a file less than 200MB");
+      //   return;
+      // }
+      let upload: tus.Upload;
+      let mediaId = "";
+      const comp: FeedCreatePostComponent = this;
+      const options = {
+        endpoint: this.backendApi._makeRequestURL(environment.uploadVideoHostname, BackendRoutes.RoutePathUploadVideo),
+        chunkSize: 50 * 1024 * 1024, // Required a minimum chunk size of 5MB, here we use 50MB.
+        // resume: true,
```
maebeam, line 266: Do we need this?

src/app/feed/feed-create-post/feed-create-post.component.ts
```diff
@@ -234,8 +244,64 @@ export class FeedCreatePostComponent implements OnInit {
...
+    if (!file) {
+      return;
+    }
+
+    if (!file.type || (!file.type.startsWith("image/") && !file.type.startsWith("video/"))) {
+      this.globalVars._alertError("File selected does not have an image or video file type.");
+      return;
+    }
+    if (file.type.startsWith("video/")) {
+      // if (file.size > 200 * (1024 * 1024)) {
```
maebeam, line 256: Do we need this?

src/app/feed/feed-create-post/feed-create-post.component.html
```diff
@@ -41,6 +41,29 @@
...
+      >
+        <iframe
+          [src]="postVideoSrc | sanitizeVideoUrl"
+          allow="accelerometer; gyroscope; autoplay; encrypted-media; picture-in-picture;"
+          allowfullscreen
+          class="feed-post__video"
+        >
+        </iframe>
+      </div>
+      <!--        <cloudflare-stream-->
```
maebeam, line 62: Do we need this?

maebeam (approved):
Looks great! Just some commented out code I'm wondering if we need

src/app/feed/feed-create-post/feed-create-post.component.ts
```diff
@@ -234,8 +244,64 @@ export class FeedCreatePostComponent implements OnInit {
...
+      if (this.videoStreamInterval != null) {
+        clearInterval(this.videoStreamInterval);
+      }
+      this.postVideoSrc = null;
+      this.readyToStream = false;
+      upload = new tus.Upload(file, options);
+      upload.start();
       return;
     }
     if (file.size > 15 * (1024 * 1024)) {
```
redpartyhat, line 307: nit: I think it would improve readability if we moved all the image processing stuff into its own if statement.  I'm imagining something like:
```
if (image) {}
else if (video) {}
else { error } 
```

src/app/backend-api.service.ts
```diff
@@ -2125,6 +2136,17 @@ export class BackendApiService {
     });
   }
 
+  UploadVideo(endpoint: string, PublicKeyBase58Check: string, FileSize: number): Observable<any> {
```
redpartyhat, line 2139: I don't actually see where this is used.  It looks like the route path is used by tus.Upload() but this function isn't actually used.  Is it needed?

lazynina: yeah we just use the routepath - I'll delete this.

src/app/feed/feed-create-post/feed-create-post.component.html
lazynina: no - I was testing out whether I preferred the cloudflare-stream component or an iframe. I think the iframe is better long term as videos are uploaded to other locations. Don't want to dirty up the codebase trying to figure out if we should use this component or an iframe. We may need to write a utility that tests out whether or not an iframe is going to work for the URL provided.

src/app/feed/feed-create-post/feed-create-post.component.ts
lazynina: no I'll specify what other options are available and where to look for the documentation on them.

redpartyhat (approved):

PR #465, opened 2021-09-27
Cosmetic handling of pending NFTs

iPaulPro:
This PR relies on https://github.com/deso-protocol/backend/pull/153

### DIFF
--- a/src/app/creator-profile-page/creator-profile-nfts/creator-profile-nfts.component.ts
+++ b/src/app/creator-profile-page/creator-profile-nfts/creator-profile-nfts.component.ts
@@ -122,12 +122,14 @@ export class CreatorProfileNftsComponent implements OnInit {
   }
 
   getNFTs(isForSale: boolean | null = null): Subscription {
+    const isPending = isForSale === null;
     return this.backendApi
       .GetNFTsForUser(
         this.globalVars.localNode,
         this.profile.PublicKeyBase58Check,
         this.globalVars.loggedInUser?.PublicKeyBase58Check,
-        isForSale
+        isForSale,
+        isPending
       )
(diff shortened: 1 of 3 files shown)


src/app/creator-profile-page/creator-profile-nfts/creator-profile-nfts.component.ts
```diff
@@ -122,12 +122,14 @@ export class CreatorProfileNftsComponent implements OnInit {
   }
 
   getNFTs(isForSale: boolean | null = null): Subscription {
+    const isPending = isForSale === null;
```
lazynina, line 125: `isForSale` is only null when the activeTab is `My Gallery`. This would make us fetch only pending NFTs on the `My Gallery` view which I don't believe is desired.  Can you clarify when we're supposed to fetch pending NFTs or what you're trying to accomplish with this line?

iPaulPro: > `isForSale` is only null when the activeTab is `My Gallery`. This would make us fetch only pending NFTs on the `My Gallery` view which I don't believe is desired. Can you clarify when we're supposed to fetch pending NFTs or what you're trying to accomplish with this line?

Apologies, you're right; this is doing exactly opposite of what I intended. 🤦‍♂️  Apologies. Just trying to filter pending NFTs from the Gallery.

I'm unable to build backend locally so I wasn't able to test this. Will push a fix when I'm able to get back to the code.

lazynina: You should be able to build it locally now - let me know if you're still having issues.

PR #468, opened 2021-09-29
Ln/twitch embed

lazynina:


redpartyhat:
In the create post component, it's not clear that you can add Twitch URLs.  We're running out of space in the placeholder text but it would be nice to communicate this somehow:
image

PR #472, opened 2021-09-30
Click-here Pointer in Tutorial

amoghgopadi:
The Click-here Pointer is shown in 

1. Step 4 after user buys their own coin - 
2. At the end of Step 2 even after user sells others coin

Hence the proposed change after which the click here pointer will be shown only in the beginning of Step 2 for the user to sell others coin

Reference - @tijno quick PR Challenge #1 - https://bitclout.com/posts/c055d0046ef8f02ca129b8fd7137c603a858564a657507cc7d1e331954e7dead?feedTab=Following

### DIFF
--- a/src/app/wallet/wallet.component.html
+++ b/src/app/wallet/wallet.component.html
@@ -252,7 +252,7 @@
     </div>
   </div>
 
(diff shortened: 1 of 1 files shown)


maebeam:
Thanks for the fix!

PR #473, opened 2021-09-30
add admin node fee management components

lazynina:


src/app/admin/admin-node-fees/admin-node-add-fee/admin-node-add-fees.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    SwalHelper.fire({
+      target: this.globalVars.getTargetComponentSelector(),
+      html: `${
+        (this.updatingAllTxns && " WARNING: this will overwrite all existing transaction fees you have set.\n") || ""
+      } Add fee of ${this.feeAmount} DESO for ${
+        this.selectedCreator?.Username || this.selectedCreator.PublicKeyBase58Check
+      }
+        for all${(!this.updatingAllTxns && " " + this.txnType) || ""} transactions.`,
```
redpartyhat, line 57: I think in general, we want to stay away from really complex in-line statements like this.  Especially interpolation like, "{(this.updatingAllTxns && " WARNING: this will overwrite all existing transaction fees you have set.\n") || "" }" that mixes logic with string creation.  I would prefer more verbose code that is easier to parse vs. more succinct code that is harder to parse.  @maebeam, @diamondhands0, I'd be curious to get your opinions on this so we can all be on the same page?

lazynina: yeah sounds good - I was messing around with what I wanted here. I don't intend for it to look this funky. There's be a clear if statement that just says `if updatingAllTxns -> stringA, else  stringB`

redpartyhat: Sounds good!

redpartyhat:
Let's add some text or something to make it more clear that the creator selected in the search bar is the *recipient* of the fees, not the one who is paying the fee.  It might be confusing for a n00b node admin.
image

redpartyhat:
I played around with everything on n0_test and everything worked as expected.  This is looking sweet, awesome work @lazynina!

redpartyhat (approved):

carry2web:
How to reset the node fees or remove a node fee for a single transaction, e.g. for Derived Key should be zero?

PR #485, opened 2021-10-16
show node fee info in sidebar

tijno:
Now that node fees are live, and some nodes use Level 4 permissions, I think the frontend should make it super simple for users to check the node fees. 

Please review whether this is an appropriate way of doing this.

* When a tx has multiple fee earners, it will sum the fees.
* It shows the top fee per Tx as a single line

20211016-GD8Upufu

* When hovering it shows all fees sorted descending

20211016-Guo08BkN

cc @lazynina


maebeam:
Seems like a great idea. Thanks tijn

PR #515, opened 2021-11-22
Add node source to posts

maebeam:


tijno:
CleanShot 2021-11-24 at 14 42 56@2x

working well @maebeam

lazynina (approved):
:+1:

iPaulPro:
Can you explain why use this centralized approach with IDs rather than just use an open standard? Would have been great to discuss this in a DIP.

maebeam:
We care about minimizing storage on chain, making it easy to update URLs/node names/owners, and protecting users from sketchy links. This solution accomplishes all these things.

It was discussed here: https://github.com/deso-protocol/dips/discussions/3#discussioncomment-1018697

PR #524, opened 2021-12-01
wait for profile comp transaction to broadcast before signing and submitting UpdateProfile transaction

lazynina:


src/app/backend-api.service.ts
```diff
@@ -1251,7 +1251,40 @@ export class BackendApiService {
       NewStakeMultipleBasisPoints,
       IsHidden,
       MinFeeRateNanosPerKB,
-    });
+    }).pipe(
+      map((res) => {
```
maebeam, line 1255: What is this code doing? I see that it checks for the txn but then what?

lazynina: If we have a `CompProfileCreatoinTxnHashHex`, poll the `get-txn` endpoint until `TxnFound` is `true`. Once found, return the response from UpdateProfile

maebeam: I could be misreading this code because it's on github and I can't read indentations but I'm not sure I understand how this accomplishes that

lazynina: I rewrote it with observables. it works as follows:

1. Construct the transaction using the update-profile endpoint
2. If the response has a `CompProfileCreationTxnHashHex`, then start polling every half second for the transaction for a minute, take the first instance where `TxnFound` is `true` and pipe the response from the update-profile request thru.
3. If the response does not have a `CompProfileCreationTxnHashHex`, pipe the response from the update-profile request thru immediately
4. sign and submit the transaction by using the piped response from steps 2 or 3.

maebeam (approved):
Observables make more sense to me ✅

PR #526, opened 2021-12-03
NFT Transfer Notifications

lazynina:
Fix console error when there are no transaction fees on this node.

image


maebeam (approved):

maebeam:
Thanks for fixing the fees error

PR #533, opened 2021-12-15
simplify loops in GetMessages that handles encryption/decryption

lazynina:


maebeam (approved):

src/app/backend-api.service.ts
```diff
@@ -1343,12 +1339,11 @@ export class BackendApiService {
...
           .pipe(
             map((decrypted) => {
-              for (const threads of res.OrderedContactsWithMessages) {
-                for (const message of threads.Messages) {
-                  message.DecryptedText = decrypted.decryptedHexes[message.EncryptedText];
-                }
-              }
-
+              res.OrderedContactsWithMessages.forEach((threads) =>
+                threads.messages.forEach(
```
AeonSw4n, line 1343: Shouldn't this be `threads.Messages` ?

AeonSw4n (changes requested):

src/app/backend-api.service.ts
maebeam: Great catch. Same error above too

AeonSw4n (approved):

PR #552, opened 2022-01-19
Fix blank screen on safari desktop with small window size

tijno:
On desktop (macOs) safari shows blank screen around 991 - 977 pixels.

CleanShot 2022-01-19 at 16 08 48

This was the simplest fix I could come up with. Better approach maybe in CSS - but problem is difference between browsers on whether the scrollbar is counted.

@superzordon let me know if you want me to PR this to diamondapp as well.


AdityaChaudhary0005:
such a waste of energy and time...

PR #560, opened 2022-02-03
Posts/Users pages are blank when user not logged in

tijno:
Issue: When a visitor hits a /posts/ or /u/ subpage but is not logged in - nothing shows. 
Expected: show post content / user profile page even when user not logged in.
Cause: Requesting metadata of undefined causes error.

CleanShot 2022-02-03 at 09 44 46@2x

This additional checks makes sure verification / node.deso.org metadata is only requested if loggedInUserPublicKey is ok.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -114,7 +114,7 @@ export class AppComponent implements OnInit {
 
     return zip(
       this.backendApi.GetUsersStateless(this.globalVars.localNode, [loggedInUserPublicKey], false),
-      environment.verificationEndpointHostname && loggedInUserPublicKey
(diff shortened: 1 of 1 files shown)


tijno:
thx to @iPaulPro for flagging this issue up. All nodes frontend forked nodes & DiamondApp affected.

cc @superzordon

tijno:
actually looks like this is not the cause!

tijno:
Aligned now with DiamondApp.

https://github.com/diamond-app/frontend/blob/main/src/app/app.component.ts#L119

lazynina (approved):

PR #561, opened 2022-02-03
Add support for Pearl node

piratekev:


lazynina (changes requested):
@piratekev - do we need to allow connect-src to *.pearl.app? Or is image-src sufficient? If it is needed, please provide an explanation

piratekev:
> @piratekev - do we need to allow connect-src to *.pearl.app? Or is image-src sufficient? If it is needed, please provide an explanation

We don't need connect-src - my mistake. That's a change for running the frontend against our node.

Removed and updated. Thanks.

lazynina (approved):

PR #564, opened 2022-02-28
display media content in replies

AdityaChaudhary0005:
bcz why not ?
before -------------------------------------------------------------------------------------------- after
image


lazynina (approved):
This resolve displaying the media in comments

PR #566, opened 2022-03-03
Use local storage to determine max post length

iPaulPro:
Instead of hardcoding the maximum post length, this change stores and reads the limit from local storage.

The benefit is that it serves the power user while changing nothing for the average user. Many users go to alternative nodes _just_ to post longer. This skews the node source metrics and interrupts the user flow unnecessarily.

For context, with BitClout+ I must completely replace the submit-post transaction handling in order to increase the max length. This is undesirable for many reasons, including it's tendency to break when any post-related changes are made to frontend.

I didn't set an upper technical limit, but perhaps there should be one hardcoded that tells the user they've set the limit too high? I believe this limit is determined by the maximum transaction size; is that correct? I've tested as many as 15k characters without issue, and was told the transaction size limit is 20kb.

### DIFF
--- a/src/app/backend-api.service.ts
+++ b/src/app/backend-api.service.ts
@@ -505,6 +505,9 @@ export class BackendApiService {
   // Messaging V3 default key name.
   DefaultKey = "default-key";
 
+  // Store the maximum characters allowed per post in localStorage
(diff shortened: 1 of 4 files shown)


tijno:
👍 great idea

PR #580, opened 2022-06-18
Add support for Piñata IPFS gateways

mvanhalen:
Add support to host pinned IPFS files via Piñata gateways. These gateways will show IPFS content faster and without rate limits. So we can use IPFS as storage solution for NFTz.me and others as well. Will also greatly improve rendering of existing IPFS NFT projects like spookies and pixelpirates (those have ratelimits now via cloudflare ipfs)

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -64,7 +64,8 @@ header @html Content-Security-Policy "
     *.arweave.net
     *.pearl.app
     *.twimg.com
-    cloudflare-ipfs.com;
+    cloudflare-ipfs.com
+    *.mypinata.cloud;
   font-src 'self'
     https://fonts.googleapis.com
     https://fonts.gstatic.com


seanslater:
Is this really required @mvanhalen ? 

Cloudflare's IPFS server will pick up anything from Pinata. The beauty of IPFS.

mvanhalen:
@seanslater yeah. Cloudflare and most public ipfs gateways are rate limited so collections will get gateway or timeout errors when showing a few images. In a grid for example. You can see it when checking spookies on nftz and start scrolling for instance.

Piñata offers a better experience can can be used by nft creators directly also.

namelessperson0:
Cloudflare isn't well respected by Tor and other privacy projects https://blog.torproject.org/trouble-cloudflare

mvanhalen:
Any feedback on this @lazynina @diamondhands0 ? We can put this live with cloudflare urls instead of chain of (stored on the blockchain). Only other nodes will have a degraded experience (rate limited and slower). Or need to get there own gateway an logic. On nftz we´ll use the piñata urls instead

lazynina (approved):

PR #589, opened 2022-08-04
Ran formatter

DeSoDog:


lazynina (commented):
Before we merge this in, let's add the prettier CI so we automatically check that new code conforms to these standard: https://github.com/marketplace/actions/prettier-action

lazynina (approved):

PR #602, opened 2022-09-22
Request Default Key generation if logged in user doesn't have one yet

lazynina:
Prompt user to create default key for messaging - this is required to get metamask users to be able to decrypt messages


brootle:
@lazynina @diamondhands0 when I saw this pop up today I thought my account was hacked, great job 👍

lazynina:
Gotta keep you on your toes @brootle - this is necessary to get messaging working for metamask users. But it will also be required for all users for group chats

PR #605, opened 2022-10-04
T/update iframe api handlers

DeSoDog:


src/app/backend-api.service.ts
```diff
@@ -1401,6 +1422,28 @@ export class BackendApiService {
         ),
       })
       .pipe(
+        tap({
+          next: (response: any) => {
+            console.log(response);
```
lazynina, line 1427: Remove console.log please
```suggestion
```

src/app/backend-api.service.ts
```diff
@@ -963,7 +964,27 @@ export class BackendApiService {
...
-            switchMap((encrypted) => {
+            tap({
+              next: (response: any) => {
+                if (response?.requestDerivedCookieWithEncryptedSeed) {
+                  this.identityService.launch('/log-in', {
+                    hideJumio: true,
+                    requestDerivedCookieWithEncryptedSeed:
+                      !!response.requestDerivedCookieWithEncryptedSeed,
+                  });
+                  return throwError(response); // end the execution of this call with an error
```
lazynina, line 975: Can we wait for the callback from the window API and then proceed if successful instead of throwing an error. Basically only throw the error if we don't have the cookie after the window API is launched?

Applies to all pipes below

src/app/identity.service.ts
```diff
@@ -208,7 +211,21 @@ export class IdentityService {
     encryptedSeedHex: string;
     transactionHex: string;
   }): Observable<any> {
-    return this.send('sign', payload);
+    return this.send('sign', payload).pipe(
```
lazynina, line 214: I think we also need to apply this  pipe to signETH and burn handlers above

lazynina: disregard this - endpoints won't throw errors

PR #629, opened 2023-02-09
Update video uploads to use livepeer client

superzordon:


lazynina (approved):

jsonpreet:
Add API endpoint for get-video-status with (Size, Resolution, and duration) due to account restriction and API access, I can't able to fetch asset details.

